PR #4, opened 2025-04-19
Add the PIIX4 SMBus driver

Steve-Tech:
Hi, I've ported the PIIX4 I2C/SMBus driver from the Linux kernel.

This protocol is commonly used on AMD motherboards (no idea about Intel, but I'm assuming they use i801), and could allow querying of RAM temperature, and configuring RGB.

~~I couldn't figure out a way to auto detect the chip's address, or verify that the address given is a PIIX4 chip.~~ (Edit: I figured things out)
For security I've just locked the address down to `0x0B00` and `0x0B20`, since OpenRGB seems to only use those addresses too. Let me know if you've got a better solution!
> Analysis of many AMD boards has shown that AMD SMBus controllers have two adapters with fixed I/O spaces at 0x0B00 and 0x0B20 - [`i2c_smbus_piix4.cpp`](https://gitlab.com/CalcProgrammer1/OpenRGB/-/blob/489d05a13f08d8fbe36747d771192be34956b612/i2c_smbus/i2c_smbus_piix4.cpp#L318)

The quick, block data, and process calls are also untested, since I didn't know how or have a way to test them.

I've got some badly written example code to query, disable, and re-enable the temperature sensors of DDR5 RAM using it's SPD5118 chip.

<details>
<summary>Python example code</summary>
<br>

```python
import ctypes
from ctypes import wintypes
import typing
from time import sleep

pawniolib = ctypes.OleDLL("C:\\Program Files\\PawnIO\\PawnIOLib.dll")

pawniolib.pawnio_version.argtypes = [ctypes.POINTER(wintypes.ULONG)]
pawniolib.pawnio_open.argtypes = [ctypes.POINTER(wintypes.HANDLE)]
pawniolib.pawnio_load.argtypes = [
    wintypes.HANDLE,
    ctypes.POINTER(ctypes.c_ubyte),
    ctypes.c_size_t,
]
pawniolib.pawnio_execute.argtypes = [
    wintypes.HANDLE,
    ctypes.c_char_p,
    ctypes.POINTER(ctypes.c_ulonglong),
    ctypes.c_size_t,
    ctypes.POINTER(ctypes.c_ulonglong),
    ctypes.c_size_t,
    ctypes.POINTER(ctypes.c_size_t),
]
pawniolib.pawnio_close.argtypes = [wintypes.HANDLE]


def pawnio_version() -> str:
    version = wintypes.ULONG()
    pawniolib.pawnio_version(ctypes.byref(version))
    major, minor, patch = (
        version.value >> 16,
        (version.value >> 8) & 0xFF,
        version.value & 0xFF,
    )
    return f"{major}.{minor}.{patch}"


def pawnio_open() -> wintypes.HANDLE:
    handle = wintypes.HANDLE()
    pawniolib.pawnio_open(ctypes.byref(handle))
    return handle


def pawnio_load(handle: wintypes.HANDLE, filepath: str) -> None:
    with open(filepath, "rb") as file:
        blob = file.read()
    size = len(blob)
    blob_array = (ctypes.c_ubyte * size)(*blob)
    pawniolib.pawnio_load(handle, blob_array, size)


def pawnio_execute(
    handle: wintypes.HANDLE,
    function: str,
    in_data: typing.Iterable,
    out_size: int,
    in_size: int | None = None,
) -> tuple[int | tuple[int]]:
    function_bytes = function.encode("utf-8")
    in_size = in_size if in_size is not None else len(in_data)
    in_array = (ctypes.c_ulonglong * in_size)(*in_data)
    out_array = (ctypes.c_ulonglong * out_size)()
    return_size = ctypes.c_size_t()

    pawniolib.pawnio_execute(
        handle,
        function_bytes,
        in_array,
        in_size,
        out_array,
        out_size,
        ctypes.byref(return_size),
    )

    return (return_size.value, tuple(out_array))


def pawnio_close(handle: wintypes.HANDLE):
    pawniolib.pawnio_close(handle)


def pawnio_ioctl(ioctl, size, *args):
    args = bytes(args)
    resp = pawnio_execute(
        handle, ioctl, args, size
    )
    return resp[1]


SPD5118_REG_TYPE            : typing.Final = 0x00	# MR0:MR1
SPD5118_REG_REVISION        : typing.Final = 0x02	# MR2
SPD5118_REG_VENDOR          : typing.Final = 0x03	# MR3:MR4
SPD5118_REG_CAPABILITY      : typing.Final = 0x05	# MR5
SPD5118_REG_I2C_LEGACY_MODE : typing.Final = 0x0B	# MR11
SPD5118_REG_TEMP_CLR        : typing.Final = 0x13	# MR19
SPD5118_REG_ERROR_CLR       : typing.Final = 0x14	# MR20
SPD5118_REG_TEMP_CONFIG     : typing.Final = 0x1A	# MR26
SPD5118_REG_TEMP_MAX        : typing.Final = 0x1c	# MR28:MR29
SPD5118_REG_TEMP_MIN        : typing.Final = 0x1e	# MR30:MR31
SPD5118_REG_TEMP_CRIT       : typing.Final = 0x20	# MR32:MR33
SPD5118_REG_TEMP_LCRIT      : typing.Final = 0x22	# MR34:MR35
SPD5118_REG_TEMP            : typing.Final = 0x31	# MR49:MR50
SPD5118_REG_TEMP_STATUS     : typing.Final = 0x33	# MR51

def get_temp():
    resp = pawnio_ioctl("ioctl_piix4_read_word_data", 1, 0x50, SPD5118_REG_TEMP)
    temp = (resp[0] >> 2)
    print(temp * 0.25, 'C')

def disable_temp(val):
    pawnio_ioctl("ioctl_piix4_write_byte_data", 1, 0x50, SPD5118_REG_TEMP_CONFIG, val)

def get_temp_config():
    temp_cfg = pawnio_ioctl("ioctl_piix4_read_byte_data", 1, 0x50, SPD5118_REG_TEMP_CONFIG)
    print(f"0x{temp_cfg[0]:x}")


print(pawnio_version())

handle = pawnio_open()
pawnio_load(handle, "I2cPIIX4.amx")

get_temp()

get_temp_config()

disable_temp(1)

get_temp_config()

get_temp()

disable_temp(0)
# JEDEC Standard No. 300-5B.01 Page 105
# If this bit is set to ‘1’ and then reset to ‘0’, the host must wait minimum of tINIT before accessing samples on the thermal sensor.
# tINIT = 10ms
# Although 10ms doesn't seem to be enough.
sleep(50/1000)

get_temp()

pawnio_close(handle)
```

</details>

I've also written a basic implementation for this module for OpenRGB: https://gitlab.com/Steve-Tech/OpenRGB/-/tree/pawnio_piix4.

For myself and everyone's reference:
- The best AMD chipset datasheet I could find: https://www.amd.com/content/dam/amd/en/documents/archived-tech-docs/programmer-references/45482.pdf
- The newest usable AMD chipset datasheet I could find: https://www.amd.com/content/dam/amd/en/documents/archived-tech-docs/programmer-references/55072_AMD_Family_15h_Models_70h-7Fh_BKDG.pdf
- The code this is based off: https://github.com/torvalds/linux/blob/v6.14/drivers/i2c/busses/i2c-piix4.c

~~Also, I'm not too sure how Windows works with I2C/SMBus, but would it be possible to implement as a PawnIO native? Then Windows should have this already implemented, and be able to handle resource contention properly.~~ Seems Windows doesn't use SMBus internally.


namazso:
Hi,

I'm not very familiar with these chips, but looking through some datasheets and the linux code, it seems that SB800 and better can be guessed fairly well from VID/DID of its PCI device, and SMBus base addresses can be read from PCI config space of the device. Could you implement that VID/DID filtering logic, and sanity check that the received SMBus base addresses are the expected ones? I think that should be good enough to at least make sure you can't just poke some random port, and hardly anyone cares about not supporting decades old hardware.

Steve-Tech:
Hi, thanks for your feedback!

That was my understanding too, but I just wasn't sure if there was an easy way to query the PCI topology to find the bus and device numbers. Although looking at some results on [linux-hardware.org](https://linux-hardware.org), it seems like AMD's 'FCH SMBus Controller' is always at `00:14.0`, so I might just validate it's PCI config before poking.

Steve-Tech:
So it seems for SB800 and above, you have to read the PMIO in MMIO for the SMBus base address, and I'm not sure how to do that in PawnIO. The Linux kernel only reads `SMBBA` from the PCI config for the older chips, and seems to use MMIO for SB800. So for now, I've just checked for the existence of a specific SMBus controller, and hoped the address is `0x0B00`.

I've been looking though this datasheet: https://www.amd.com/content/dam/amd/en/documents/archived-tech-docs/programmer-references/45482.pdf - but let me know if you've found a newer one!

namazso:
Oh huh, indeed. In fact it seems that that would be the only reasonable one to implement since the chipse before are also ancient. You can try [these](https://github.com/namazso/PawnIO.Modules/blob/main/include/native.inc#L37-L48) functions to read from MMIO

Steve-Tech:
Thanks for the suggestion for the functions to use! I got MMIO working, so the base address gets properly validated now.

If you're interested, I've also written a basic implementation for this module for OpenRGB: https://gitlab.com/Steve-Tech/OpenRGB/-/tree/pawnio_piix4.

namazso:
Signed: [SmbusPIIX4.zip](https://github.com/user-attachments/files/19950707/SmbusPIIX4.zip)

PR #8, opened 2025-05-03
Add AMDReset to read the S5_RESET_STATUS on AMD systems

Steve-Tech:
Hi, my day job is a computer technician, and being able to read the reset/shutdown reason seems like it *might* be useful, although I kinda doubt it ever will.

This commit was what inspired this: https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?id=b343579de7b250101e4ed6c354b4c1fa986973a7

But proper documentation is usually in volume 5 in [AMD's Processor Programming Reference](https://www.amd.com/en/search/documentation/hub.html#q=PPR&sortCriteria=%40amd_release_date%20descending&f-amd_document_type=Programmer%20References). I've been using [this one](https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/programmer-references/57228.zip), because it was the first one I clicked on.

I have also considered adding an ioctl to write `0x6` (warm reset) or `0xE` (cold reset) to IO Port `0xCF9`, but that seems slightly too dangerous. (For reference, a cold reset will 'place system in S5 state for 3 to 5 seconds').

Thanks,
Steve

### DIFF
--- a/AMDReset.p
+++ b/AMDReset.p
@@ -1,0 +1,143 @@
+//  PawnIO Modules - Modules for various hardware to be used with PawnIO.
+//  Copyright (C) 2025  Steve-Tech <me@stevetech.au>
+//
+//  This library is free software; you can redistribute it and/or
+//  modify it under the terms of the GNU Lesser General Public
+//  License as published by the Free Software Foundation; either
+//  version 2.1 of the License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


namazso:
I'm surprised this works, I didn't know MmMapIoSpace will add / remove offsets as needed, that's why it was documented to only take aligned addresses.

namazso:
Signed build: [AMDReset.zip](https://github.com/user-attachments/files/20022861/AMDReset.zip)

Steve-Tech:
I've written a simple program here, for anyone that wants to use it. https://gist.github.com/Steve-Tech/100bf6958156fb7c4c5df122202c2d48

PR #9, opened 2025-05-03
PIIX4: Don't reset to main address when calling ioctl_piix4_port_sel with -1

Steve-Tech:
Hi, we can't put `-1` in with the other ports, because it'll reset `piix4_smba` back to the main port, and screw with anyone calling it with `-1` while using the aux port.

Thanks,
Steve

### DIFF
--- a/SmbusPIIX4.p
+++ b/SmbusPIIX4.p
@@ -416,7 +416,10 @@ public NTSTATUS:ioctl_piix4_port_sel(in[], in_size, out[], out_size) {
     new NTSTATUS:status = STATUS_SUCCESS;
 
     switch (new_port) {
-    case -1, 0, 2, 3, 4:
+    case -1:
+        if (old_addr == addresses[0])
+            status = piix4_port_sel(new_port, old_port);
+    case 0, 2, 3, 4:
         {
             piix4_smba = addresses[0];
             status = piix4_port_sel(new_port, old_port);
@@ -432,9 +435,8 @@ public NTSTATUS:ioctl_piix4_port_sel(in[], in_size, out[], out_size) {
         return status;
 
     if (old_addr == addresses[1])
-        out[0] = 1;
-    else
-        out[0] = old_port;
+        old_port = 1;
+    out[0] = old_port;
 
     return status;
 }


namazso:
I'm not sure I understand the reasoning here. How would any other scenario not screw someone, like if they're using Port 1 and call it with Port 0?

With this PR the only way to write -1 in the selector from when you're currently using the aux port is to first switch to 0, then to -1, which is a very confusing usage pattern.

namazso:
Thinking about it a bit more, it seems that Linux's port numbering is just silly. Digging through the blame, I found [this](https://github.com/torvalds/linux/commit/725d2e3facfb51f8925da3cadce5380b5ea85d09). It seems they had primary and auxilary as 0 and 1 pre-SB800, and when the primary port multiplexer was added they had to go with a skipped slot to keep stability. I think what we should do instead is have one ioctl for port selection, and another for selecting between the primary and auxilary base, probably how it should've always been. This also makes it easily extendible if they were to add a multiplexer on auxilary too.

Steve-Tech:
> I'm not sure I understand the reasoning here

Port 1 is actually on a separate base address, 0, 2, 3, 4, are not. Case 0, 2, 3, 4 resets `piix4_smba` back to `addresses[0]` in case it had been changed. So if someone was using port 1 (the aux port on it's own address), and called port_sel with `-1`, then it resets the base address back to the main one.

Or in code:
```pawn
// Assume port 1 is selected, `piix4_smba == addresses[1];`
// We want to double check for debugging or something so we call ioctl_piix4_port_sel with -1

    case -1, 0, 2, 3, 4:
        {
            piix4_smba = addresses[0];
            // Oh no, now `piix4_smba == addresses[0];`
            // This would be fine if port 1 wasn't selected
            status = piix4_port_sel(new_port, old_port);
        }
    case 1:
        // Port 1 is the aux port, we just change the base address
        piix4_smba = addresses[1];
```

> it seems that Linux's port numbering is just silly

Oh the port numbering comes from the datasheet:
For example in the 15h one:
image
Or the 19h one (It's missing the actual smbus registers though):
image

Also since ports 3 and 4 aren't in the modern datasheets, should we just remove them?

namazso:
Oh, okay, I see now, -1 is solely used for learning the currently selected port, I missed how it doesn't actually mutate any device state.

Steve-Tech:
I can split that functionality into another ioctl if you'd like, I guess it makes more sense to keep everything more specialised.

namazso:
Can you check if https://github.com/namazso/PawnIO.Modules/commit/fa574b3ff55349ed985400d6097d0af9198c4918 and https://github.com/namazso/PawnIO.Modules/commit/975cb1dee93cc99245155d2c60dd692081144818 works for this goal? It should sanely handle every combination of (primary, aux) -> (primary, aux, -1)

namazso:
Actually no I messed it up (https://github.com/namazso/PawnIO.Modules/commit/e1b16cd8f1004d17aa31f5a0ad4f5fc1dec84851). I really should've just made this into a fresh branch and a new PR.

Steve-Tech:
Yep, I have confirmed that works!

PR #11, opened 2025-05-17
PIIX4 Optimisations

Steve-Tech:
Hi, this fixes the issues that Peter Berendi brought up in: [OpenRGB!2833](https://gitlab.com/CalcProgrammer1/OpenRGB/-/merge_requests/2833#note_2483982487)

> I can only comment on the piix4 driver. While it should basically work, there are apparently some stability and performance issues with it.
>
> - Host controller busy status is first tested in piix4_transaction, after other registers are already written. The datasheet says no other registers should be accessed while the busy bit is set.
> - The microsleep function is using KeDelayExecutionThread, its accuracy is limited by the system clock period, so it would actually wait for a duration between 250 and 750 μs. Half a millisecond isn't much if there is a single transaction, but let's suppose you have 4 ram sticks with 12 rgb leds each, that's 144 registers to change. It adds up, visibly slowing down software effects.
>
> See [!2702](https://github.com/CalcProgrammer1/OpenRGB/-/merge_requests/2702) for code that addresses these issues (for WinRing0), accessing registers in the way described in the documentation.

In my testing, this also seems to reduce the time of reading a byte 1000 times, from ~15ms per read, to ~0.15ms per read.

### DIFF
--- a/SmbusPIIX4.p
+++ b/SmbusPIIX4.p
@@ -85,9 +85,62 @@
 new addresses[] = [0x0B00, 0x0B20];
 new piix4_smba = 0x0B00;
 
+new tsc_freq;
+
+// Gets the TSC frequency in Hz
+NTSTATUS:get_tsc_freq()
+{
+    // This method is only supported on AMD 17h (Ryzen) and later processors
+    new vendor[4];
+    cpuid(0, 0, vendor);
+    if (!is_amd(vendor))
+        return STATUS_NOT_SUPPORTED;
+
+    new procinfo[4];
+    cpuid(1, 0, procinfo);
+
+    new family = ((procinfo[0] & 0x0FF00000) >> 20) + ((procinfo[0] & 0x0F00) >> 8);
+    if (family < 0x17)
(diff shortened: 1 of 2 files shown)


namazso:
There's only one bus, we probably shouldn't clog it up with a tight loop of I/O instructions. You could use `microsleep2` to at least spin a few micros so that the other cores get to use the bus a bit.

Steve-Tech:
Done! Is there a rule against sleeping for more than 50us with `microsleep2`, `microsleep` seems way too slow for 100us.

Currently at ~0.16ms per read.

namazso:
There's no rule, it's just what Microsoft recommends: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-kestallexecutionprocessor

It has no consequence other than burning some cycles like the previous one, we just don't want to spam I/O while we do that.

namazso:
Thanks! I just released [0.1.0](https://github.com/namazso/PawnIO.Modules/releases/tag/0.1.0) with signed blobs for all modules.

PR #12, opened 2025-05-18
PIIX4: Fix the AUX port

Steve-Tech:
Sorry it seems I forgot to push a commit in my last PR.

The AUX port never seems to set bit 1 (0x02) if there is no response, causing it to timeout instead of erroring early.


Steve-Tech:
Just found out and thought I might document it: This issue only affects my laptop, not my desktop. This PR doesn't have any adverse effects on either systems.

So the AUX port wasn't completely broken all along.

PR #13, opened 2025-05-21
SMBus Timing Improvements

Steve-Tech:
Hi,

i801 now uses `microsleep2` for more accurate sleeping, and only delays for the theoretical minimum amount of time needed based on the number of bytes being transferred. i801 style controllers operate at a max of 100khz, and there isn't a way to read (or change?) the speed.

PIIX4 also now waits for the theoretical minimum amount of time based on clock speed and bytes transferred. My laptop (where I did the original benchmarking) seemed to run at ~392KHz, while my desktop runs at ~93KHz; so it's no longer hard coded to somewhat arbitrary values, and will adjust based on clock speed. Until SMBus 3.0, the max speed was 100KHz.

Both drivers now delay in the loop for one clock cycle, since that seems reasonable.

Here's the most relevant SMBus spec: https://www.smbus.org/specs/smbus20.pdf
Here's the latest SMBus spec: https://www.smbus.org/specs/SMBus_3_3_1_20241020.pdf

Thanks,
Steve

### DIFF
--- a/SmbusPIIX4.p
+++ b/SmbusPIIX4.p
@@ -48,7 +48,6 @@
 #define I2C_SMBUS_I2C_BLOCK_DATA    8
 
 /* Other settings */
-#define MAX_TIMEOUT		500
 #define  ENABLE_INT9	0
 
 /* PIIX4 constants */
@@ -82,6 +81,9 @@
 #define SMBSLVEVT	(0x0A + piix4_smba)
 #define SMBSLVDAT	(0x0C + piix4_smba)
 #define SMBTIMING	(0x0E + piix4_smba)
+
+// A 32 byte block read at 10MHz is 32.5ms, 64ms should be plenty
+#define MAX_TIMEOUT 64
 
 new addresses[] = [0x0B00, 0x0B20];
 new piix4_smba = 0x0B00;
--- a/SmbusPIIX4.p
+++ b/SmbusPIIX4.p
@@ -81,6 +81,7 @@
 #define SMBSHDWCMD	(0x09 + piix4_smba)
 #define SMBSLVEVT	(0x0A + piix4_smba)
 #define SMBSLVDAT	(0x0C + piix4_smba)
+#define SMBTIMING	(0x0E + piix4_smba)
 
 new addresses[] = [0x0B00, 0x0B20];
 new piix4_smba = 0x0B00;
@@ -184,20 +185,26 @@ NTSTATUS:piix4_busy_check()
     return STATUS_SUCCESS;
 }
 
-NTSTATUS:piix4_transaction()
+NTSTATUS:piix4_transaction(size)
 {
     new NTSTATUS:status = STATUS_SUCCESS;
     new temp;
+    new timing = io_in_byte(SMBTIMING);
 
     /* start the transaction by setting bit 6 */
     io_out_byte(SMBHSTCNT, io_in_byte(SMBHSTCNT) | 0x040);
 
     // Don't wait more than MAX_TIMEOUT ms for the transaction to complete
     new deadline = get_tick_count() + MAX_TIMEOUT;
-    // Testing shows we're usually done in 120-160us
-    microsleep2(100);
+
+    // 11 bits minimum (start + slave address + rd/wr + ack + stop)
+    // 9 bits per byte (byte + ack)
+    // From datasheet: 'Frequency = 66Mhz/(SmBusTiming * 4)' (we flip the division for the period)
+    microsleep2(((11 + (9 * size)) * timing * 4) / 66);
     do {
-        microsleep2(20);
+        // Only check for result once per clock cycle
+        // Also allows for 1 cycle of processing time
+        microsleep2((timing * 4) / 66);
         temp = io_in_byte(SMBHSTSTS);
     } while ((get_tick_count() < deadline) && (temp & 0x01));
 
@@ -235,20 +242,20 @@ NTSTATUS:piix4_transaction()
     return status;
 }
 
-NTSTATUS:piix4_access_simple(addr, read_write, command, size, in, &out)
+NTSTATUS:piix4_access_simple(addr, read_write, command, hstcmd, in, &out)
 {
-    new NTSTATUS:status;
+    new NTSTATUS:status, protocol, size = hstcmd;
 
     status = piix4_busy_check();
     if (!NT_SUCCESS(status))
         return status;
 
-    switch (size) {
+    switch (hstcmd) {
     case I2C_SMBUS_QUICK:
         {
             io_out_byte(SMBHSTADD, 
                         (addr << 1) | read_write);
-            size = PIIX4_QUICK;
+            protocol = PIIX4_QUICK;
         }
     case I2C_SMBUS_BYTE:
         {
@@ -256,7 +263,7 @@ NTSTATUS:piix4_access_simple(addr, read_write, command, size, in, &out)
                         (addr << 1) | read_write);
             if (read_write == I2C_SMBUS_WRITE)
                 io_out_byte(SMBHSTCMD, command);
-            size = PIIX4_BYTE;
+            protocol = PIIX4_BYTE;
         }
     case I2C_SMBUS_BYTE_DATA:
         {
@@ -265,7 +272,7 @@ NTSTATUS:piix4_access_simple(addr, read_write, command, size, in, &out)
             io_out_byte(SMBHSTCMD, command);
             if (read_write == I2C_SMBUS_WRITE)
                 io_out_byte(SMBHSTDAT0, in);
-            size = PIIX4_BYTE_DATA;
+            protocol = PIIX4_BYTE_DATA;
         }
     case I2C_SMBUS_WORD_DATA:
         {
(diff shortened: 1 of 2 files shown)


SmbusPIIX4.p
```diff
@@ -184,20 +185,26 @@ NTSTATUS:piix4_busy_check()
...
+
+    // 11 bits minimum (start + slave address + rd/wr + ack + stop)
+    // 9 bits per byte (byte + ack)
+    // From datasheet: 'Frequency = 66Mhz/(SmBusTiming * 4)' (we flip the division for the period)
+    microsleep2(((11 + (9 * size)) * timing * 4) / 66);
     do {
-        microsleep2(20);
+        // Only check for result once per clock cycle
+        // Also allows for 1 cycle of processing time
+        microsleep2((timing * 4) / 66);
```
PeterBerendi, line 207: We know exactly how long the transaction would take, there's no need for sleep here.

SmbusPIIX4.p
```diff
@@ -235,28 +242,28 @@ NTSTATUS:piix4_transaction()
     return status;
 }
 
-NTSTATUS:piix4_access_simple(addr, read_write, command, size, in, &out)
+NTSTATUS:piix4_access_simple(addr, read_write, command, hstcmd, in, &out)
 {
-    new NTSTATUS:status;
+    new NTSTATUS:status, protocol, size = hstcmd;
```
PeterBerendi, line 247: `size = hstcmd + read_write` because read transactions send `addr` twice

SmbusPIIX4.p
Steve-Tech: Idk, in my tests it seems to take around 2 to 9 cycles longer (interestingly though it's quite exact at 392KHz). Also if something happens we don't want to be spamming the bus until the timeout ends. I guess we could shorten the timeout.

SmbusPIIX4.p
Steve-Tech: Ahh good catch, although it seems to only be for the `*_data` calls?

SmbusPIIX4.p
Steve-Tech: Seems consistently 2 cycles longer than expected now.

PeterBerendi: Add 2 cycles to the initial delay then.

I couldn't find an exact timing specification for PIIX4 and the I2C/SMBus standard specifies minimum times only. 

Some I2C host adapters stretch the start/stop signals to 2 cycles, others don't. Looks like this one does.

SmbusPIIX4.p
PeterBerendi: Yes. Do you think anyone would notice if reading the temperature from an LM77 took 300 μs instead of 200?

SmbusPIIX4.p
Steve-Tech: I don't really want to remove the sleep, since if something goes wrong we're hammering the IO bus until the timeout expires. My understanding also (actually based on i801) is clock stretching can be used down to 10kHz, so in the worst case scenario we're hammering the bus for 32.5ms (or actually 64ms just in case).

@namazso what are your thoughts on this?

namazso: I think your sleep should be ok: If the timings are correct, this amount of sleep should get us ever so slightly off-phase (due to computations we do in between) from the clock so that we check right after a result can possibly arrive. This should be better than sleeping a small but arbitrary amount, and probably smarter than a 100% busy spin. Not sleeping would only really be significantly beneficial if the timing was very wrong in the first place. If our timings were as confident as @PeterBerendi 's initial comment suggested, we wouldn't even need the loop - why bother checking it again if we know it must've already finished?

You can probably benchmark a busy version and the sleeping version to see how off we are. If the sleeps are right it should be maybe a microsecond or two, if they're wrong it should average at half the sleep duration and we should reconsider picking sleep length.

However, in either case this sleep would only account for less than 5% of the total time, so maybe it isn't worth worrying that much about it. WinRing0 which we're replacing already has a ~1-2us "sleep" between every spin due to the inherent cost of an ioctl. If we wanted to beat it by more we could add a batched ioctl instead.

PeterBerendi: I didn't mean we don't need the loop, I meant we might not have to sleep.

We can't be 100% sure that nothing ever happens that would delay the transaction.

How about increasing sleep duration?
```
  delay = 0;
  do {
    microsleep2(delay++);
    temp = io_in_byte(SMBHSTSTS);
  } while ((get_tick_count() < deadline) && (temp & 0x01));
```

namazso: In theory that should be worse than sleeping exactly one cycle for the earlier mentioned reason, but that would be my suggestion as well if benchmark comes back with the conclusion that we’re off by an unknown amount.

Steve-Tech: I've done some benchmarking:

Current PR

- ioctl_piix4_read_byte: 251.26 ns
- ioctl_piix4_read_byte_data: 473.1075 ns
- ioctl_piix4_write_byte_data: 361.5722 ns
- ioctl_piix4_read_word_data: 569.9704 ns

No sleeps

- ioctl_piix4_read_byte: 248.4869 ns
- ioctl_piix4_read_byte_data: 467.3571 ns
- ioctl_piix4_write_byte_data: 350.456 ns
- ioctl_piix4_read_word_data: 568.8731 ns

No loop sleep

- ioctl_piix4_read_byte: 249.5951 ns
- ioctl_piix4_read_byte_data: 467.4164 ns
- ioctl_piix4_write_byte_data: 351.0448 ns
- ioctl_piix4_read_word_data: 569.9961999999999 ns

No loop sleep with 2 extra cycle inital sleep

- ioctl_piix4_read_byte: 260.0201 ns
- ioctl_piix4_read_byte_data: 467.2328 ns
- ioctl_piix4_write_byte_data: 357.3159 ns
- ioctl_piix4_read_word_data: 569.701 ns

Increasing delay

- ioctl_piix4_read_byte: 251.7911 ns
- ioctl_piix4_read_byte_data: 470.9051 ns
- ioctl_piix4_write_byte_data: 357.40259999999995 ns
- ioctl_piix4_read_word_data: 575.9831999999999 ns

namazso: As chart:

image

Interesting results, though they look odd enough that stddev would be also useful. Increasing performs worse the more it waits (as expected), but it seems that there's no off-phase timing benefit from checking every cycle because half-cycle sometimes performs better which it shouldn't if that was the case.

My recommendation would be increasing up until half-cycle then staying at that frequency. This should eliminate instances when we're just unlucky with timing while still not hogging the bus 100% the time. This should keep us around the inherent jitter the requests seem to have.

PeterBerendi: I'm getting 335 μs (+/- 2 μs) for write_byte_data using WinRing0 on a Ryzen 7 5800X. The code is basically no sleep, but the loop condition is a bit different. https://gitlab.com/peterberendi/OpenRGB/-/blob/kingstonfury_work/i2c_smbus/i2c_smbus_piix4.cpp?ref_type=heads#L141

I was hoping that fiddling with the sleep duration would make PawnIO as fast as WinRing0 or even faster but that doesn't seem to be the case.

namazso: I wonder if it might come down to hardware or benchmarking differences, because it really should just be overall faster. (Just a syscall costs like 300ns from the start!) Unfortunately I don't have a PIIX4 to just benchmark both side by side.

Thinking about it, the difference between your number and the no sleep version is 16 μs, which is suspiciously close to 20 μs aka 2 cycles at 100 kHz, the difference @Steve-Tech noticed in behavior.

namazso: The time it should take on the wire with the bonus bits is `((10+(9*2)+2)*176*4)/66` = 320μs. That'd mean you had only 13μs overhead best case. The linked code intentionally sleeps 10μs, leaving 3μs remaining. Assuming the `in` instructions are free (unlikely) and you only pay the ~1μs kernel round trip fee, you do 6 calls to WinRing0 resulting in -3μs overhead remaining.

I think your device actually just doesn't have the two bonus bits, that gets more realistic WinRing0 overhead numbers.

Steve-Tech: >  The linked code intentionally sleeps 10us

That's only if `delay_timer` is set right? I had assumed that code was running without any sleep.

I was still using Python though for my benchmark, since I was more meaning to compare different methods than an actual speed test. I'll get around to writing a C version eventually (possibly next weekend), and hopefully comparing it to WinRing0 on my system.

namazso: > That's only if delay_timer is set right?

Indeed, I guess in reality it probably waits less on the other branch. If we assume it to be 0, then it's 13μs for 6 `in`/`out`-s, which would be very close (~2.2μs per WinRing0 call, close to the ~2.3μs I benched on a PawnIO-implemented `in` ioctl). But then the question is where the remaining time goes in this module, we do barely any computation (certainly not 20μs worth) and a PawnIO-implemented `in` benches only 100ns or so slower so its also not the overheads. I still find the fact that the difference is so close to 20μs suspicious though.

SmbusPIIX4.p
```diff
@@ -184,20 +185,26 @@ NTSTATUS:piix4_busy_check()
...
+    // 9 bits per byte (byte + ack)
+    // From datasheet: 'Frequency = 66Mhz/(SmBusTiming * 4)' (we flip the division for the period)
+    microsleep2(((11 + (9 * size)) * timing * 4) / 66);
     do {
-        microsleep2(20);
+        // Only check for result once per clock cycle
+        // Also allows for 1 cycle of processing time
+        microsleep2((timing * 4) / 66);
         temp = io_in_byte(SMBHSTSTS);
     } while ((get_tick_count() < deadline) && (temp & 0x01));
```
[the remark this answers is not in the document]
Steve-Tech: I've made some changes in my last couple commits to attempt this.

SmbusPIIX4.p
namazso: Thanks, very interesting dataset. Odd how no sleep's timings are not within a few σ of current pr, I'd expect oversleeping to be the single difference there.

I'm also surprised at the stddev values in general: assuming our starting point compared to a 100 kHz clock cycle is uniformly distributed, we should have at least sqrt(10^2/12) = 2.88 μs standard deviation just from the start. Maybe this is just the side effect of testing in a tight loop. However, that means in non-benchmark scenarios the variance from the sleeps is less than just what we inherently get from not being synced up to the clock.

Steve-Tech: > I'm also surprised at the stddev values in general

Yeah, the stddev values are also probably a little broken since it's the deviation between 100 batches of 100 runs. I could try timing individual runs, but I have a feeling `std::chrono::high_resolution_clock` probably won't measure things properly.

Thought I might add the theoretical minimum time (or the initial sleep time + stop bit + read start bit) to the graph as a reference:
image

PR #15, opened 2025-07-06
SMBus: Implement a standard ioctl for SMBus modules

Steve-Tech:
Hi,

As mentioned in [OpenRGB !2885](https://gitlab.com/CalcProgrammer1/OpenRGB/-/merge_requests/2885#note_2600407736), this PR implements `ioctl_smbus_xfer` for PIIX4 and i801.
~~I have marked it as a draft as I intend on getting feedback and implement it for i801 too.~~

~~I am also wondering if the old ioctls should be kept too? Linux does have both [implementations](https://www.kernel.org/doc/html/latest/i2c/dev-interface.html#full-interface-description), and could make future ports easier; but I also can see the simplicity viewpoint.~~

Thanks,
Steve

### DIFF
--- a/SmbusI801.p
+++ b/SmbusI801.p
@@ -632,325 +632,196 @@ unlock:
     return status;
 }
 
-/// SMBus quick write.
+/// Identify the SMBus controller.
 ///
-/// The SMBus Read/Write Quick protocol (SMBQuick) is typically used to control
-/// simple devices using a device-specific binary command (for example, ON and OFF).
-/// Command values are not used by this protocol and thus only a single element
-/// (at offset 0) can be specified in the field definition.
-///
-/// @param in [0] = Address, [1] = Command
-/// @param in_size Must be 2
-/// @param out Unused
-/// @param out_size Unused
+/// @param out [0] = Type of the SMBus controller, [1] = I/O Base address, [2] = PCI Identifiers
+/// @param out_size Must be 3
 /// @return An NTSTATUS
-/// @warning You should acquire the "\BaseNamedObjects\Access_SMBUS.HTP.Method" mutant before calling this
-forward NTSTATUS:ioctl_i801_write_quick(in[], in_size, out[], out_size);
-public NTSTATUS:ioctl_i801_write_quick(in[], in_size, out[], out_size) {
-    if (in_size < 2)
+forward NTSTATUS:ioctl_identity(in[], in_size, out[], out_size);
+public NTSTATUS:ioctl_identity(in[], in_size, out[], out_size) {
+    if (out_size < 3)
         return STATUS_BUFFER_TOO_SMALL;
 
-    new address = in[0];
-    new command = in[1];
+    new NTSTATUS:status;
 
-    new unused;
+    out[0] = CHAR4_CONST('i', '8', '0', '1');
 
-    return i801_access_simple(address, command, 0, I2C_SMBUS_QUICK, 0, unused);
+    out[1] = i801_smba;
+
+    // Read the PCI vendor/device ID
+    new pci_ids;
+    status = pci_config_read_dword(pci_addr[0], pci_addr[1], pci_addr[2], 0x00, pci_ids);
+    if (!NT_SUCCESS(status))
+        return status;
+
+    // Read the PCI subsystem vendor/device ID
+    new pci_subsys_ids;
+    status = pci_config_read_dword(pci_addr[0], pci_addr[1], pci_addr[2], 0x2C, pci_subsys_ids);
+    if (!NT_SUCCESS(status))
+        return status;
+
+    out[2] = pci_ids | (pci_subsys_ids << 32);
+
+    return STATUS_SUCCESS;
 }
 
-/// SMBus byte receive.
+/// Set the SMBus clock frequency.
 ///
-/// The SMBus Send/Receive Byte protocol (SMBSendReceive) transfers a single
-/// byte of data. Like Read/Write Quick, command values are not used by this
-/// protocol and thus only a single element (at offset 0) can be specified in
-/// the field definition.
-///
-/// @param in [0] = Address
+/// @param in [0] = Frequency in Hz or -1 for no change
 /// @param in_size Must be 1
-/// @param out [0] = Data
+/// @param out [0] = Previous frequency in Hz
 /// @param out_size Must be 1
+/// @note The i801 SMBus controller use a fixed clock frequency of 100kHz.
 /// @return An NTSTATUS
-/// @warning You should acquire the "\BaseNamedObjects\Access_SMBUS.HTP.Method" mutant before calling this
-forward NTSTATUS:ioctl_i801_read_byte(in[], in_size, out[], out_size);
-public NTSTATUS:ioctl_i801_read_byte(in[], in_size, out[], out_size) {
+forward NTSTATUS:ioctl_clock_freq(in[], in_size, out[], out_size);
+public NTSTATUS:ioctl_clock_freq(in[], in_size, out[], out_size) {
+    if (out_size < 1)
+        return STATUS_BUFFER_TOO_SMALL;
     if (in_size < 1)
         return STATUS_BUFFER_TOO_SMALL;
-    if (out_size < 1)
-        return STATUS_BUFFER_TOO_SMALL;
 
-    new address = in[0];
+    out[0] = 100000;
 
-    return i801_access_simple(address, I2C_SMBUS_READ, 0, I2C_SMBUS_BYTE, 0, out[0]);
+    new new_freq = in[0];
+    if (new_freq != -1) {
+        return STATUS_NOT_SUPPORTED;
+    }
+
+    return STATUS_SUCCESS;
 }
 
-/// SMBus byte send.
+/// SMBus transfer.
 ///
-/// The SMBus Send/Receive Byte protocol (SMBSendReceive) transfers a single
-/// byte of data. Like Read/Write Quick, command values are not used by this
-/// protocol and thus only a single element (at offset 0) can be specified in
-/// the field definition.
+/// Performs a transfer of data over the SMBus using the specified command.
+/// I2C_SMBUS_QUICK (protocol 0) only requires the address and read/write parameters, command must be left as 0.
+/// I2C_SMBUS_BYTE (1), I2C_SMBUS_BYTE_DATA (2), and I2C_SMBUS_WORD_DATA (3) require the address, read/write, command, and data (write only) parameters.
+/// I2C_SMBUS_PROC_CALL (4) requires the address, command, and data parameters.
+/// I2C_SMBUS_BLOCK_DATA (5) requires the address, read/write, command, data as length, and array data parameters.
+/// I2C_SMBUS_BLOCK_PROC_CALL (7) requires the address, command, data as length, and array data parameters.
 ///
-/// @param in [0] = Address, [1] = Data
-/// @param in_size Must be 2
-/// @param out Unused
-/// @param out_size Unused
-/// @return An NTSTATUS
-/// @warning You should acquire the "\BaseNamedObjects\Access_SMBUS.HTP.Method" mutant before calling this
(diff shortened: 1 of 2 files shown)


CalcProgrammer1:
Linux doesn't implement the other functions in the kernel, in that page you linked it says the functions are provided by a library.  I would remove them from the kernel drivers.

namazso:
> I would be fine implementing i2c_smbus_xfer/i2c_xfer as done in OpenRGB or taking it a step further and implementing i2c_slave, i2c_smbus, and i2c_rdwr ioctls in the PawnIO drivers and using them pretty much how the Linux ioctls are called in i2c_smbus_linux.cpp.

After taking a look at https://github.com/torvalds/linux/blob/master/drivers/i2c/i2c-dev.c it appears all I2C_SLAVE does is storing the value in kernel memory to just pass it every time when converting a `write()` or `I2C_SMBUS` into a `smbus_i2c_xfer` call. Don't think we need completely unnecessarily introduced state on the PawnIO side, especially not due to the bad decisions Linux made a long time ago.

Regarding `i2c_xfer`, that needs `I2C_FUNC_I2C` on Linux, which neither of these have, as apparently SMBus controllers usually don't support raw I2C.

CalcProgrammer1:
Yeah, i2c_xfer should be implemented only for controllers that support raw i2c, we don't need to worry about implementing that one for now.

Steve-Tech:
> Regarding `i2c_xfer` ... apparently SMBus controllers usually don't support raw I2C.

i801 seems to have an I2C mode and I2C read command, but I don't know anything about it. I could potentially try hooking up an Arduino or something to my SMBus later, but don't think it's important for now.

SmbusPIIX4.p
```diff
@@ -85,6 +85,7 @@
 // A 32 byte block read at 10MHz is 32.5ms, 64ms should be plenty
 #define MAX_TIMEOUT 64
 
+new pci_addr[3] = [0x00, 0x14, 0x0];
```
namazso, line 88: These should be constants:

```
#define PIIX_PCI_BUS      0x00
#define PIIX_PCI_DEVICE   0x14
#define PIIX_PCI_FUNCTION 0x00
```

SmbusPIIX4.p
```diff
@@ -493,17 +494,34 @@ public NTSTATUS:ioctl_piix4_port_sel(in[], in_size, out[], out_size) {
...
+    if (out_size < 4)
         return STATUS_BUFFER_TOO_SMALL;
-    
-    out[0] = piix4_smba;
+
+    new NTSTATUS:status;
+
+    new name{8} = "PIIX4";
+    // This isn't great, but it gets around 'warning 229: mixing packed and unpacked array indexing or array assignment'
+    out[0] = name{0} << 56 | name{1} << 48 | name{2} << 40 | name{3} << 32 |
```
namazso, line 511: Same as the other, but you can just implement the other versions (1 to 8) of that macro as well.

SmbusI801.p
```diff
@@ -632,17 +632,34 @@ unlock:
...
+public NTSTATUS:ioctl_identity(in[], in_size, out[], out_size) {
+    if (out_size < 4)
         return STATUS_BUFFER_TOO_SMALL;
-    
-    out[0] = i801_smba;
+
+    new NTSTATUS:status;
+
+    new name{8} = "i801";
+    // This isn't great, but it gets around 'warning 229: mixing packed and unpacked array indexing or array assignment'
```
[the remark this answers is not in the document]
Steve-Tech: Yeah I thought about that, but I also wanted to keep it readable. I'll implement `CHAR8_CONST` and use that.

SmbusPIIX4.p
namazso: Oh I guess you’d might not have dealt with multichar constants in C on x86… Usually you write them backwards if you want a certain pattern in memory.

CalcProgrammer1:
How do you use the test version of the driver?  I have Windows in test mode and installed the unsigned version of the driver.  If I use `'C:\Program Files\PawnIO\PawnIOUtil.exe' test .\SmbusI801.bin` it shows in DebugView:

```
[PawnIO] Signature check result: C000000D
[PawnIO] debug_print: SmbusI801 start
```

I added the debug_print to the main() function.  However, when I copy these .bin files into OpenRGB and try to load them from i2c_smbus_pawnio.cpp, it fails with:

```
[ERROR:] Failed to load SmbusI801.bin, PawnIO initialization aborted
[ERROR:] Failed to load SmbusPIIX4.bin, PawnIO initialization aborted
[ERROR:] Failed to load SmbusPIIX4.bin, PawnIO initialization aborted
```

I removed the PCI ID checks so it tries to load both, and I would expect SmbusI801 to work on this PC.  However, it just says failed to load which means `pawnio_load` returned an invalid status.  Nothing shows in DebugView when opening OpenRGB, and yes I opened it as administrator.

namazso:
Where are you getting the .bin files from? Compiling with pawncc should result in a .amx file, which you sign to get a .bin file. The Unrestricted version simply ignores the result of the signature check. `PawnIOUtil sign` should be able to sign with any RSA2048 key. Alternatively, you can just prefix your .amx with 4 null bytes to get an invalid-signed .bin.

An additional ways to test stuff is `PawnIOUtil interactive` which lets you test interactively and just call anything with any params. The [source code](https://github.com/namazso/PawnIOLib/blob/main/PawnIOUtil/PawnIOUtil.cpp) is probably the best reference for now to see how it does things.

CalcProgrammer1:
I just renamed the .amx to .bin, if I want to load it without modifying the OpenRGB side can I just append 4 null bytes?

CalcProgrammer1:
```
[Info:] PawnIO initialized
[Info:] [ResourceManager] Registering I2C interface: @��ޤ Device FFFFFFFF:FFFFFFFF Subsystem: FFFFFFFF:FFFFFFFF
[ERROR:] Failed to load SmbusPIIX4.bin, PawnIO initialization aborted
[ERROR:] Failed to load SmbusPIIX4.bin, PawnIO initialization aborted
```

Nice, that worked.  The corrupt name and ID is expected, now I'm going to fill them in using the new ioctl.

CalcProgrammer1:
I'm not sure if I'm doing this right:
```
    /*-----------------------------------------------------*\
    | Get bus information                                   |
    \*-----------------------------------------------------*/
    const SIZE_T    in_size         = 1;
    ULONG64         in[in_size]     = {0};
    const SIZE_T    out_size        = 4;
    ULONG64         out[out_size];
    SIZE_T          return_size;
    HRESULT         status;

    status = pawnio_execute(handle, "ioctl_identify", in, in_size, out, out_size, &return_size);

    LOG_TRACE( "i2c_smbus_pawnio: ioctl_identify return status %d", status);

    //if(!status)
    {
        LOG_TRACE( "i2c_smbus_pawnio: ioctl_identify data: %16X %16X %16X %16X", out[0], out[1], out[2], out[3]);

        this->pci_vendor            = (out[2] & 0x000000000000FFFF);
        this->pci_device            = (out[2] & 0x00000000FFFF0000) >> 16;
        this->pci_subsystem_vendor  = (out[3] & 0x000000000000FFFF);
        this->pci_subsystem_device  = (out[4] & 0x00000000FFFF0000) >> 16;

        char name_str[9];
        name_str[0]                 = (out[0] & 0x00000000000000FF);
        name_str[1]                 = (out[0] & 0x000000000000FF00) >> 8;
        name_str[2]                 = (out[0] & 0x0000000000FF0000) >> 16;
        name_str[3]                 = (out[0] & 0x00000000FF000000) >> 24;
        name_str[4]                 = (out[0] & 0x000000FF00000000) >> 32;
        name_str[5]                 = (out[0] & 0x0000FF0000000000) >> 40;
        name_str[6]                 = (out[0] & 0x00FF000000000000) >> 48;
        name_str[7]                 = (out[0] & 0xFF00000000000000) >> 56;
        name_str[8]                 = 0;

        strncpy( this->device_name, name_str, 512 );
    }
```

It doesn't look like device name is being filled properly and the PCI IDs don't look right (Intel is 8086 IIRC).  What value should a successful status be?

```
[Trace:] i2c_smbus_pawnio: ioctl_identify return status -2147024894
[Trace:] i2c_smbus_pawnio: ioctl_identify data:                0                1         EBB94450         F4F65038
[Info:] [ResourceManager] Registering I2C interface:  Device 4450:EBB9 Subsystem: 5038:BF7C
```

namazso:
Glad you got it working!

Regarding this:

> What value should a successful status be?

HRESULT should be tested with the `SUCCEEDED` or `FAILED` macros. Although in reality by inspecting PawnIOLib code we know it'll always be system facility which can only report a single success status (0) so testing against S_OK will also work, and so would using FormatMessage if you wanted to give some more descriptive error text.

Steve-Tech:
> It doesn't look like device name is being filled properly

Sorry that was my fault, it should be fixed now.

Also I've packed the PCI IDs in my last commit, so the ioctl_identity struct is now:

```c
struct SMBUSIdentity {
    char Name[8];
    uint64_t CurrentBase;
    uint16_t VendorId;
    uint16_t DeviceId;
    uint16_t SubsysVendorId;
    uint16_t SubsysDeviceId;
};
```

CalcProgrammer1:
Is there a reason in[4] and in[5...8] are two separate interfaces for passing data?

I think it would be better (and line up with the Linux implementation more) if in[4...end] were just byte-packed "data", array or otherwise.  For byte and word data, it would be the same as an array of 1 or 2 bytes.  For actual variable-width block transfers, the first byte would be the length (same as is done in Linux).  Same would be true for output.

CalcProgrammer1:
I'm thinking about adding functions get_byte_le and set_byte_le in core.inc.  The goal of these functions would be to directly access byte arrays packed into cells bytewise.

For instance, if in[4...8] is a 34-byte array, instead of unpacking into 34 cells (wasting 8x the space), you would just access bytes out of the in array directly by calling `byte_val = get_byte_le(in, <0-33>) `.  Similarly, to set an individual byte in the out[0...4] array, you could just call `set_byte_le(out, <0-33>, byte_val)`.

I want to use this access strategy in my NCT6793 driver.

namazso:
I'd much prefer keeping static sized arrays, as they can be automatically bounds checked by the compiler and so only the unpacking has to be correct. Using `[]` eliminates these checks. Additionally, if you zero copy then you have to be aware that the buffer behind `src` is the same as `dst` which might be very unexpected for some.

One middle-ground alternative could be moving to symbolic subscripts and packed arrays, but the later are dodgy at best in the compiler and kind of an abomination made only to not waste space on strings:

```
#define ARG[.id, .arr{16}]

test(a{16}) {
    return a{1};
}

test2(a[ARG]) {
    return test(a.arr);
}

main() {
    new v[ARG];
    v.arr{1} = 2;
    printf("%d\n", test2(v));
}
```

SmbusI801.p
```diff
@@ -632,325 +632,196 @@ unlock:
...
-forward NTSTATUS:ioctl_i801_write_word_data(in[], in_size, out[], out_size);
-public NTSTATUS:ioctl_i801_write_word_data(in[], in_size, out[], out_size) {
-    if (in_size < 3)
+    if (in_size < 1)
         return STATUS_BUFFER_TOO_SMALL;
 
-    new address = in[0];
-    new command = in[1];
-    new data = in[2];
+    out[0] = 100000;
```
namazso, line 683: Actually, the current code here suffers from the issue I just mentioned in my previous comment: `out` is backed by the same buffer a s `in`. The only reason this works by chance is that the out buffer starts 32 bytes earlier because the input ioctl has the symbol name. If you wrote `out[4]` here then the next read would just read out `100000`.

namazso:
@Steve-Tech I cherry picked your macros onto main, and squashed the rest of your commits. I also added two changes, could you review them?

CalcProgrammer1:
> @Steve-Tech I cherry picked your macros onto main, and squashed the rest of your commits. I also added two changes, could you review them?

Can you also review and cherry-pick my new bytewise access macros?  I can squash and rebase my MR if you approve them.

Steve-Tech:
@namazso I've tested what I can, and I don't see any issues.

PR #16, opened 2025-07-19
AMD Family 10

namazso:


namazso:
Worst case it'll just BSOD if broken, but I think this was fixed by https://github.com/namazso/PawnIO.Modules/commit/62cb3068b00b7b1c499faf5b2c0683dea482b19d#diff-231711505eaf644b68fc855af4ad7226821719063ea76bc6c5a2562360dd75fe

PR #17, opened 2025-07-29
NCT6793D: Timing improvements

Steve-Tech:
Hi, this adds some timing improvements for NCT6793D, as was done with the other drivers.

This is untested, so I've marked it as a draft for now.

@CalcProgrammer1, it would be great if you could test and review this!


CalcProgrammer1:
Looks like a detection fail:

```
[Debug:] [ENE (ASUS Aura) SMBus Controller] Bus 02 is a motherboard and the subvendor matches the one for ASUS, looking for a device at 0x4E
[Debug:] [ENE SMBus] looking for devices at 0x4E...
[Debug:] [ENE SMBus] Detected an I2C device at address 4E, testing register range
[Verbose:] [ENE SMBus] Detection failed testing register A6.  Expected 06, got FFFFFFFF.
```

PR #18, opened 2025-08-15
LpcCrOSEC: Implement byte packing

Steve-Tech:
Hi,

This implements byte packing to the LpcCrOSEC module, like we did with the SMBus modules.

It also adds Doxygen comments to the ioctls, as well as making some variable name changes.

Thanks,
Steve

### DIFF
--- a/LpcCrOSEC.p
+++ b/LpcCrOSEC.p
@@ -104,7 +104,7 @@ NTSTATUS:wait_for_ec(const status_addr, const timeout_usec) {
 	return STATUS_TIMEOUT; /* Timeout */
 }
 
-NTSTATUS:ec_command_lpc_3(command, version, ec_out_data[], ec_out_size, ec_in_data[], ec_in_size, &result) {
+NTSTATUS:ec_command_lpc_3(command, version, ec_out_data[EC_LPC_HOST_PACKET_SIZE], ec_out_size, ec_in_data[EC_LPC_HOST_PACKET_SIZE], ec_in_size, &result) {
     new csum = 0;
     const rq_size = 1 + 1 + 2 + 1 + 1 + 2;
     const rs_size = 1 + 1 + 2 + 2 + 2;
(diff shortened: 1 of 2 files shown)


LpcCrOSEC.p
```diff
@@ -271,33 +271,82 @@ NTSTATUS:comm_init_lpc() {
...
-        new version = in[2];
-        new in_size_offset = 3;
-        new ec_in_size = in_size - in_size_offset;
-        new out_size_offset = 1;
-        new ec_out_size = out_size - out_size_offset;
-
-        return ec_command_lpc_3(command, version, in, ec_in_size, in_size_offset, out, ec_out_size);
+        unpack_bytes_le(in, in_data, ec_out_size, 4 * 8, 0);
+
+        status = ec_command_lpc_3(command, version, in_data, ec_out_size, out_data, ec_in_size, result);
```
namazso, line 309: All usages of ec_out_size and ec_in_size in this function are swapped. Since it's consistently wrong it'd work out, but very confusing.

LpcCrOSEC.p
```diff
@@ -271,33 +271,82 @@ NTSTATUS:comm_init_lpc() {
...
     if (out_size < 1)
         return STATUS_BUFFER_TOO_SMALL;
 
+    new NTSTATUS:status;
     new offset = in[0] & 0xFFFF;
+    new bytes = in[1];
+    new out_data[EC_MEMMAP_SIZE];
 
-    return ec_readmem_lpc(offset, out_size, out);
+    if (out_size < _ceil_div(bytes, 8))
```
namazso, line 340: exploitable overflow. if I set in[1] to 0x7FFFFFFFFFFFFFFF, this condition will pass but pack_bytes_le will attempt packing that amount of bytes, overwriting heap.

Steve-Tech: Oh woops, I'm guessing `==` would fix that? Or should I change pack_bytes to `out_size * 8`?

LpcCrOSEC.p
namazso: I see. Why not name the unpacked arrays ec_out and ec_in then? Really the confusing part is that something named `blah_size` doesn't contain the size of `blah`. A comment about it wouldn't hurt above the respective variables.

Steve-Tech: Done!

PR #19, opened 2025-08-31
Changes for SMBus I801 & PIIX4

Blacktempel:
Add PCICMD handling, in case I/O for SMBus is disabled.

Add `write_protection_enabled` variable and `ioctl_write_protection` to get write protection status from application.

Change fixed `I2C_SMBUS_WRITE` for `I2C_SMBUS_PROC_CALL` to dynamic `read_write`:
- DDR5 page change address is write protected, but there is a workaround where we need to do a `I2C_SMBUS_READ` for `I2C_SMBUS_PROC_CALL` and we can change the page successfully
_(see usage: https://github.com/Blacktempel/RAMSPDToolkit/blob/9b2aeab9b7637be1874520c74c9873e174fe4947/RAMSPDToolkit/SPD/DDR5AccessorBase.cs#L237)_

Also write read value from PROC_CALL into `out[0]` instead of discarding.

### DIFF
--- a/SmbusI801.p
+++ b/SmbusI801.p
@@ -68,7 +68,7 @@
 #define TCOBASE		0x050
 #define TCOCTL		0x054
 
-#define PCICMD_IOBIT	BIT(1)
+#define PCICMD_IOBIT	0x01
 
 /* Host configuration bits for SMBHSTCFG */
 #define SMBHSTCFG_HST_EN		BIT(0)
@@ -215,7 +215,6 @@ new pci_devices[] = [
 new pci_addr[3];
 new i801_smba;
 new write_protection_enabled;
-new pci_cmd_original;
 
 NTSTATUS:i801_init()
 {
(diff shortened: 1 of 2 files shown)


SmbusI801.p
```diff
@@ -705,103 +726,137 @@ DEFINE_IOCTL(ioctl_smbus_xfer) {
     new read_write = in[1];
     new command = in[2];
     new hstcmd = in[3];
-    
+
+    new pci_cmd_modified;
+
+    //PCI CMD IO not enabled
+    if (false == (pci_cmd_original & PCICMD_IOBIT))
```
namazso, line 733: I think pci_cmd_original should be fetched here instead (and probably be a local). I think you can only guarantee no others are touching this stuff while `"\BaseNamedObjects\Access_SMBUS.HTP.Method"` is held, which (according to the warning) is only while we're in this function, and not for the whole lifetime of the PawnIO module.

namazso (changes requested):
Thanks! I'll give the control flow another pass of review, but the changes make sense to me.

namazso (approved):

PR #20, opened 2025-09-09
SmbusI801: Fix size bug that can lead to excessive delays with high v…

CalcProgrammer1:
…alue device register addresses

See discussion here: https://gitlab.com/CalcProgrammer1/OpenRGB/-/merge_requests/2885#note_2737907187


Steve-Tech (approved):

namazso:
I’ll wait with the release for the PIIX4 investigation, plus LpcIO also has some bugs

PR #25, opened 2025-09-12
Simplify LpcIO for now

### LINKED ISSUE #21: LpcIO detect bug
From: https://github.com/Rem0o/FanControl.Releases/issues/3016#issuecomment-3267575645

From what I can tell, the problem is you might get a non zero value for let's say a [Winbond](https://github.com/namazso/PawnIO.Modules/blob/2c94ef0a4c7c0a8abb67cf11aecdd2fcc6da487d/LpcIO.p#L158)  but that id might not match anything relevant from the software perspective (think LHM), so you'll never check for ITE and other types.

Couldn't the 3 enter/exit be split into 3 methods, so we could check after each if we get a valid hit? 

ioctl_detect_Winbond
ioctl_detect_ITE
ioctl_detect_Smsc

Then from the app perspective:

```c#
var res = _pawnIO.Execute("ioctl_detect_Winbond");
if ( IsValid(res, out Chip chip) )
    return chip;

res = _pawnIO.Execute("ioctl_detect_ITE");
if ( IsValid(res, out chip) )
    return chip;

// ...
```
If I got the idea wrong, feel free to correct me.

Everyone in FanControl who face detection issue seem to have ITE chips, which is the 2nd detection cycle, so it adds up with this theory.

There is also this bit of code that might be relevant that is missing: https://github.com/LibreHardwareMonitor/LibreHardwareMonitor/blob/1704f8c7b733a39aeeb2b8435ae20f0e1ff31dda/LibreHardwareMonitorLib/Hardware/Motherboard/Lpc/LpcIO.cs#L573

namazso:
Fixes #21

Removes the whole Gigabyte EC stuff for now. It'd probably be worth to just re-make it from scratch, maybe as a separate module

### DIFF
--- a/LpcIO.p
+++ b/LpcIO.p
@@ -17,13 +17,6 @@
 
 #include <pawnio.inc>
 
-const _: {
-    Vendor_Unknown = 0,
-    Vendor_Winbond,
-    Vendor_IT87,
-    Vendor_Smsc
-};
-
 const CHIP_ID_REGISTER = 0x20;
 const CHIP_REVISION_REGISTER = 0x21;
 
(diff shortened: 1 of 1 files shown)


ironicmoustache:
> @namazso I tried it, and can confirm that the ITE IT8792E related fans are now showing and FanControl launch correctly without error, however the associated fans are not controllable.
> 
> Tested on a Gigabyte X570 Master with FanControl v236, .Net 8.0.
> 
> This looks promising, thanks!

Stumbled on this thread as I've been trying to keep up with the situation and am in the same boat on a Gigabyte B550 Aorus Pro AC. Case fans appear, albeit without working speed control, but it's a positive step. On an even more positive note, however, the CPU cooler fan control is now working fine.

namazso:
Thanks for testing!

@Ozzuneoj 

> am I understanding correctly that you are also the developer of PawnIO, which this .dll uses?

Yes. Ideally the main LHM will be using PawnIO soon and I won't be maintaining a fork on the long term.

@ironicmoustache 

> albeit without working speed control

Yeah, the fans are overridden by the Gigabyte EC firmware so attempting to configure them is ignored. Gigabyte boards will likely need some code for that, but the current detection was a mess. I'm hoping to find some saner or easier to implement method via datasheets, reversing, or testing.

Rem0o:
Awesome work @namazso

Ozzuneoj:
Just wanted to say, it's been three days since I switched to the LHM in [this post](https://github.com/namazso/PawnIO.Modules/pull/25#issuecomment-3286494233) and I still haven't had any problems. Everything seems to be working as expected. :)

PR #26, opened 2025-09-13
Gigabyte EC

namazso:


namazso:
If it’s not a machine you use / care about then just an RDP with admin would work, and I figure out the rest.

Otherwise, I can prepare a test binary to see how far we get, and hopefully it won’t devolve into poking at random stuff in RWEverything.

hexawyz:
Hi there
I do have Gigabyte Hardware (Intel, Z490 Vision D) as my main computer.
I won't have much time to investigate anything, but if there is something easy that you want to check, maybe I can look at it. (No promises though)

yqfQiQkH0OhC5:
> Running FaceIT AC by any chance?

@namazso I use FaceIT AC. Is there a workaround? I recently got the defender issue with winring (openrgb and fancontrol) and I saw that in the new versions they use pawnio. My MB is an Asus Z170-k.

namazso:
@yqfQiQkH0OhC5 short of uninstalling / unloading faceit, no, not for now. I guess you could write them in hopes of whitelisting, but likely the same signer also signed some vulnerable drivers at some point. So a solution from this side needs to wait for me to get another signer which will take a bit of time.

yqfQiQkH0OhC5:
@namazso Thanks for your efforts and your time. I'm looking forward to that.

simy1129:
Is there any chance that Asus motherboard e.g. TUF B850 also not fully supported by PawnIO? I had installed PawnIO and together with FanControl. But it couldn't detect any motherboard fan controllers. 
It shouldn't be windows defender issues, as I whitelisted FanControl and it works when I rollback to previous FanControl versions.

namazso:
@simy1129 It is intended to be supported. The only intentional regression is losing the ability to control MB fans on Gigabyte, and restoring that is what this PR is about. Everything else is a bug in something that is intended to work, and should be reported as such.

namazso:
@CalcProgrammer1 This is being tracked at https://github.com/namazso/PawnIO.Setup/issues/1 and a solution might be on the horizon. The complicated licensing model of PawnIO is meant to subsidize a proper signing certificate. However, I'll keep HYTE in mind if anything falls through.

namazso:
Sorry, I was a bit busy dealing with other stuff. Here's a test program based on this current branch: [untitled5.zip](https://github.com/user-attachments/files/22453890/untitled5.zip)

Run from terminal, as admin, and copy output (or you can also redirect it to a file, whatever works).

In theory it should also enable your fan control if disabled, and disable if it's enabled (like if you run it a second time).

bluearcher-bc:
[output.txt](https://github.com/user-attachments/files/22454868/output.txt)

VikingGer:
Don't know if it helps, but here's some info from my system, too:
X470 AORUS ULTRA GAMING-CF, ITE IT8686

[output.txt](https://github.com/user-attachments/files/22464261/output.txt)

namazso:
Thanks, that actually looks closer to what I'd expect. Still looks pretty wrong though, supposedly the controller was disabled and we only enabled it? Anyway, here's a newer test: [untitled5.zip](https://github.com/user-attachments/files/22476887/untitled5.zip)
This should now properly print original state, and dump with both disabled and enabled address decoding, to see if I even enable the right thing at all or we're just seeing some memory garbage.

bluearcher-bc:
Unclear if I need to run this new one twice (to disable then re-enable) but this is my output after only running the new one once.
[output.txt](https://github.com/user-attachments/files/22480945/output.txt)

Dschogo:
[output.txt](https://github.com/user-attachments/files/22548082/output.txt) taken with this test: https://github.com/namazso/PawnIO.Modules/pull/26#issuecomment-3321337554

seems to kinda work. The fans (connected to second chip) definitely respond, but wont allow me to go below 25% speed. Above 25% they follow the controls perfectly.

FanControl V241
x470 Gigabyte Aorus Gaming Ultra

KyleJ61782:
I was having the same issues, but the test `untitled5.exe` executable seems to work for me. Here are the logs for enabling and disabling fan control. Interestingly, no SuperIO dumps are collected.
[output_enabled.txt](https://github.com/user-attachments/files/22548558/output_enabled.txt)
[output_disabled.txt](https://github.com/user-attachments/files/22548559/output_disabled.txt)

bluearcher-bc:
Extremely interesting utilities from TsunamiMommy in the thread above 👀

namazso:
@Dschogo Neat, so I guess the code isn't just fundamentally flawed if it at least works for some. Maybe some chip family has incorrect decoding?

@KyleJ61782 That's very odd since it evidently worked, I wonder if my test app has some more bugs I missed. Well, it was hastily thrown together so it wouldn't be very surprising.

@bluearcher-bc I was away for a bit, but I'll look into the stuff posted here more in-depth over the weekend, for now only more hardware variance would be interesting. 

> Extremely interesting utilities from TsunamiMommy in the thread above 👀

Seems like it's mostly a port of the code here, so it's not very useful for comparing for bugs, but the dumps and reversing work of the layout can be of interest.

TsunamiMommy:
Hi guys. I see that you're having some issues with control of some of the superios. The main issue you're having with accessing the second superios is that you're entering and exiting config mode on them. The IT8883 espi to lpc bridge chips will enter config mode instead and will block access to the second superio. So in order for reliable access you cannot enter the config mode on those. At least I don't with a superio utility I built for linux based on this MR. Hopefully that helps. I also have mapping data for the smartfan stuff on the superios too. Though since I have the IT8696 and IT87952 superios I could only map them out. You can see this issue for more information. https://github.com/frankcrawford/it87/issues/64 Let me know if you guys need/want more data. I did document most of it to the best of my ability. You can use my utility on linux or port it over to pawnio if you want to do more advanced diagnostics.

ThomasHFWright:
Hey, the test fixed my pump speed! Wooo
[output.txt](https://github.com/user-attachments/files/22764254/output.txt)

I've got a GIGABYTE X870 A Elite WIFI7 Ice

bluearcher-bc:
@TsunamiMommy would your python work via WSL? I'm guessing probably not

TsunamiMommy:
@bluearcher-bc Unfortunately it wouldn't work in WSL. As WSL doesn't abstract the entire set of hardware to the virtualized linux kernel. You could map it to pawnio (or even winring0). It does what this mr does in order to expose the mmio space/ec space but gives it real time read/write capability along with the ability to export the data to another program that can decode it. A live usb with just about any distro with python installed can do the job as well. That's probably the easiest solution.

TsunamiMommy:
@cosmintnt can you give me the system configuration and control software you're using?

cosmintnt:
I did try to boot without HWinfo as startup, run the utility, but got the same outcome. 
I also use OpenRGB, but as far as I know, that's stil on winring0. For the test thou, I also stopped OpenRGB as well.
ZenTimmings is also linked to pawnio, but it was not running during the test and it does not check for fan sensors anyway.
[output_1.txt](https://github.com/user-attachments/files/22920185/output_1.txt)

TsunamiMommy:
I see. It seems that whatever is going on, something is blocking access to the second superio. Something else you can do is run my python utility on a linux live usb and capture data that way. And if your motherboard has system 7 and system 8 pumps then it stands to reason that your system also has the third chip for those two headers. Another thing you could try I guess is a full shutdown and removal of power for 10 seconds or so then start back up again.

Blacktempel:
@cosmintnt LHM and therefore FanControl supports RAM temperature since a couple of months.
Is it not working for you or did you not see the update / sensors yet ? Please let me know.
You may have to enable Memory in FanControl settings.

> No GCC installed, but I'm running HWinfo and FanControl on the same time, as I use a FC plugin to get RAM temperature from HWinfo and adjust RAM fan based on dimm's temperature. image

cosmintnt:
@Blacktempel - Thanks for the headsup, I was doing it for so long with HWinfo plug in ...

@TsunamiMommy: is there a way to chat somewhere about the topic? to avoid clutering the thread with relatively off-topic subject (like guiding me with Linux live usb, your script and trying to find what's not working). For sure we'll share possible findings in a more condensed way than replaying here back and forth.

DarkAssassinUA:
untitled5 helped me,thank you.
Motherboard: Gigabyte B500 Aorus Pro AC
[output_enabled.txt](https://github.com/user-attachments/files/22934507/output_enabled.txt)

DarkAssassinUA:
[log.txt](https://github.com/user-attachments/files/22993280/log.txt)
tested again,Gigabyte B500 Aorus Pro AC, worked.

namazso:
But then the question remains how would it have worked with the original WinRing0 implementation in LHM? it was doing pretty much exactly the same as what I did here.

namazso:
Oh, actually that's simple to explain: ECIO was added to LHM at some point, after I started porting a bunch of stuff over: https://github.com/LibreHardwareMonitor/LibreHardwareMonitor/pull/1313

Yeah I guess that's a whole different thing to cover then

TsunamiMommy:
Yup. So it did help us figure it out. Because my gp utility can access ECIO if smfi isn't supported. And I saw valid data there. That's how I came to that conclusion.

Rem0o:
Opened a [LHM PR draft](https://github.com/LibreHardwareMonitor/LibreHardwareMonitor/pull/2014)

Need all the review I can get, just put the code in there but obv I couldn't test anything. Feedback/help welcomed.

TsunamiMommy:
So, something that needs to be addressed with this code is that based on some testing and code verification I've done, intel systems _cannot_ support two simultaneous windows. So that behavior will need to be changed. I've been working with a few linux users with intel systems and the behavior from this implementation (that I've mirrored in the it87 driver), only keep the last mmio address written to the `0x98` base address register on the intel isa bridge. This by default won't work because if you try to write an additional address there it will replace the previous one and keep only that window active. There may be additional steps for skylake platforms as well, but I'm still testing it.

PR #38, opened 2025-11-17
Allow send SMU commands, read and write SMU registers

Erruar:
I have added the functionality to send SMU commands within a limited range. I have also added methods for reading and writing SMU registers within the ranges used by my application. The work has been thoroughly tested on two separate devices and everything functions as expected. These ranges will not affect the anti-virus software and should not cause any issues. This modification is essential for the proper operation of my project


Erruar:
@namazso Thank you so much for merging my changes. I hope that our cooperation will continue to be just as productive.

PR #51, opened 2026-02-26
LpcCrOSEC: Fix unexpected extra reads/writes

Steve-Tech:
See https://github.com/namazso/PawnIO.Modules/pull/46#discussion_r2857638042


Steve-Tech:
Thanks for your help!

PR #53, opened 2026-05-01
Add Intel client IMC clock ratio module (IntelIMC.p)

DevTechProfile:
# Add Intel client IMC clock ratio module (IntelIMC.p)

## What this adds

A new read-only module, `IntelIMC.p`, exposing the IMC's QCLK ratio on modern Intel client SoCs through a single IOCTL, `ioctl_read_imc_clock`. The output carries the ratio together with its reference clock mode and gear so a consumer can compute a memory-clock sensor against its own measured BCLK.

## Why it's needed

Tools like CapFrameX and LibreHardwareMonitor have computed "Memory Clock" by reading `MSR_UNCORE_PERF_STATUS` (0x621) and multiplying its uncore ratio with the bus clock. From Ice Lake onward — and especially on Meteor / Arrow / Lunar / Panther Lake — the IMC and uncore are decoupled, so that MSR no longer represents the memory clock and produces plausible-but-wrong values. Intel publishes the IMC ratio in MCHBAR-mapped MMIO registers instead, which is what this module reads.

## How it works

The module resolves the CPU via a strict CPUID allowlist and dispatches to one of two MCHBAR-relative registers, both at compile-time constant offsets:

- **Core Ultra (MTL, ARL, LNL, PTL)** → `MEMSS_PMA_CR_BIOS_DATA` at `MCHBAR + 0x13D10` (locked QCLK ratio referenced to BCLK/3, plus a Gear2/Gear4 bit).
- **Alder / Raptor Lake** → `SA_PERF_STATUS` at `MCHBAR + 0x5918` (live workpoint ratio with a BCLK vs. BCLK×4/3 reference bit).

`SA_PERF_STATUS` is deliberately not read on Core Ultra — Intel's own docs mark its `QCLK_RATIO` field as "not defined properly" there.

## Safety posture

- One IOCTL with `in_size = 0`. No user-controlled byte reaches the kernel.
- Every PCI BDF and register offset is a compile-time constant. No writes to PCI/MMIO/MSR/IO. MCHBAR enable bit is observed but never modified.
- Reads are sanity-checked: reserved bits in `MEMSS_PMA_CR_BIOS_DATA` must be zero, ratio must fall in a 16–220 band; otherwise `STATUS_NOT_SUPPORTED`.
- Allowlist excludes overlapping IDs explicitly (e.g. `0xCF` = Emerald Rapids, not Panther Lake).

## Coverage

14 CPUID models across 6 platforms, taken from `intel-perfmon/mapfile.csv`:

| Platform | Model IDs | Source |
|---|---|---|
| Meteor Lake | 0xAA, 0xAC, 0xB5 | MEMSS_PMA |
| Arrow Lake | 0xC5, 0xC6 | MEMSS_PMA |
| Lunar Lake | 0xBD | MEMSS_PMA |
| Panther Lake | 0xCC, 0xD5 | MEMSS_PMA |
| Alder Lake | 0x97, 0x9A, 0xBE | SA_PERF_STATUS |
| Raptor Lake | 0xB7, 0xBA, 0xBF | SA_PERF_STATUS |

Skylake through Comet Lake are unchanged and stay on the existing `IntelMSR` path. ICL/TGL/RKL are omitted until their IMC publishing register is confirmed.

## Validation status

No allowlisted platform has had real-hardware validation yet. Every successful return therefore sets `IMC_FLAG_EXPERIMENTAL`, and the source instructs consumers to keep the sensor hidden by default or label it experimental. Per-platform validated flags will be added (and the experimental bit cleared) as results land.

## References

- Intel Core Ultra 200H/200U CFG/MEM — `MEMSS_PMA_CR_BIOS_DATA`, MCHBAR base
- Intel 14th-gen client CFG/MEM — `SA_PERF_STATUS`
- `intel-perfmon/mapfile.csv` (V1.05 PTL, V1.17 ARL, V1.21 MTL, V1.22 LNL, V1.39 ADL/RPL)


DevTechProfile:
In case you are willing to provide a beta binary, I am able to test it with Lunar Lake, Arrow Lake, Panther Lake, and Raptor Lake.

namazso:
Hi,

You can test without signatures with the unrestricted driver:

https://github.com/namazso/PawnIO.Modules/wiki/Getting-started-with-PawnIO

I'd much prefer if this was tested on at least one of the supported platforms.

namazso:
I did a quick review on the code, I have some observations:

- Is there any argument against just exposing the whole 128 kB region for reading? It looks benign and readable without sideeffects.
- About the comment in `main`: I believe error statuses from main should travel back just fine, and a quick review of PawnIO code seems to confirm this. Are you sure it didn't work for you?
- It seems that both CPU groups that have this share a DID/VID of 8086/4600 (according to docs, but it's not on https://pci-ids.ucw.cz/ ), would it make more sense to restrict based on this instead, or could this change (or rather, remain the same while the semantics for accessing it change) at Intel's discretion?

a1ive:
> It seems that both CPU groups that have this share a DID/VID of 8086/4600 (according to docs, but it's not on [pci-ids.ucw.cz](https://pci-ids.ucw.cz/) ), would it make more sense to restrict based on this instead, or could this change (or rather, remain the same while the semantics for accessing it change) at Intel's discretion?

Even within the same generation of Core CPUs, the DID is not the same.
image

The method for reading MCHBAR has been compatible with later generations since at least Sandy Bridge, see https://github.com/memtest86plus/memtest86plus/blob/main/system/imc/x86/intel_snb.c

DevTechProfile:
Thanks for taking the time to a quick review @namazso.

### Validation status

The module is now validated on three Core Ultra platforms with HWiNFO64 + CPU-Z cross-checks and SMBIOS DIMM-speed match:

| Platform | CPU / CPUID | Module result | Cross-checked against |
|---|---|---|---|
| Panther Lake | Core Ultra X7 358H / 0xCC | ratio 64, Gear4, **IO 4,267 MHz / 8,533 MT/s** | HWiNFO Memory Clock 4,266.7 MHz, Gear 4; SMBIOS 8,533 MT/s |
| Arrow Lake | Core Ultra 9 285K / 0xC6 | ratio 117, Gear2, **QCLK 3,900 MHz / 7,800 MT/s** | HWiNFO Memory Clock 3,899.6 MHz, Gear 2; SMBIOS 7,800 MT/s; G.Skill EXPO DDR5-7800 |
| Lunar Lake | Core Ultra 7 258V / 0xBD | ratio 64, Gear4, **QCLK 2,133.3 MHz / 8,533 MT/s** | HWiNFO Memory Clock 2,133.3 MHz, Ratio 21.33x; SMBIOS 8,533 MT/s |

Full reports are in the branch _test_infrastructure_ under `Deploy/VALIDATION-{PTL,ARL,LNL}.md`. `is_platform_validated()` (IntelIMC.p:227) clears `IMC_FLAG_EXPERIMENTAL` for these three. MTL and NVL stay flagged as unvalidated.

### Point 2 — comment in `main()`

The comment was not saying that error statuses from `main` don't propagate. They do, and the module already returns `STATUS_NOT_SUPPORTED` from `main` for non-x64 / non-Intel hosts.

What I deliberately don't gate in `main` is the **CPU-model check**, and the reason is UX, not a workaround:

- "Module loaded, IOCTL returns `STATUS_NOT_SUPPORTED`" → the user knows their CPU is simply not on the allowlist.
- "Module failed to load" → could be a PawnIO driver issue, a missing dependency, or anything else.

The first is easy to tell apart from infrastructure problems; the second isn't.

The comment has been reworded to say exactly that (IntelIMC.p:726-727):

```pawn
// No model gate here; per-IOCTL allowlist returns STATUS_NOT_SUPPORTED
// so callers can distinguish "didn't load" from "model not allowlisted".
```

### Point 3 — DID/VID instead of CPUID

@a1ive already covered this with the screenshot above: DIDs differ within the same generation, so DID-based gating is more brittle than CPUID, not less.

CPUID also matches Intel's own platform classification. The allowlist in this module is cross-checked against `mapfile.csv` from Intel's public `intel-perfmon` repo (referenced at IntelIMC.p:46 and IntelIMC.p:162, mirrored under `intel-perfmon/` here for offline audit). That gives an auditable upstream source for what each model is.

The "Intel could change semantics" concern applies equally to DID and CPUID, so DID brings no extra safety — only extra maintenance per SKU.

### Point 1 — exposing the full 128 kB MCHBAR region

This is the one that's worth a longer reply. The narrow surface is intentional, and the security envelope at the top of the file lists it as a design goal (IntelIMC.p:35-41):

> - Read-only (no writes to PCI cfg / MMIO / MSR / IO).
> - No caller-controlled physical addresses or PCI BDF.
> - Fixed compile-time MCHBAR offsets only.
> - Strict CPUID allowlist; unknown models return `STATUS_NOT_SUPPORTED`.
> - Reserved-bits and ratio-range checks reject wrong-register reads.
> - MCHBAR enable bit is observed, never set.

The narrowness is what makes the safety checks meaningful:

- The reserved-bits mask in `read_memss_pma` and the `IMC_RATIO_MIN..MAX` range check work *because* the module knows which register it's reading. A wrong-stepping or wrong-platform read fails closed.
- A user-space tool can't accidentally turn an unknown MCHBAR field into a "Memory Clock" sensor.
- The audit surface for a security review stays small and stable.

A generic `read_mchbar_dword(offset)` would lose all of that. MCHBAR isn't only IMC — it also covers DMI, PCIe, IMPH, etc. Reads are side-effect-free, agreed, but exposing arbitrary architectural reserved fields weakens the defense-in-depth story: the module would no longer be self-contained in what it reads.

If you'd like a wider surface for diagnostics or future sensors, I'd suggest a *separate* IOCTL — for example `ioctl_read_mchbar_dword` — with either:

- an allowlist of permitted offset ranges, or
- a bounds check `[0, 0x20000)` plus dword alignment.

That keeps the public Memory-Clock API tight and labels the wider path explicitly as a diagnostic / research surface. Happy to fold either option into this PR if you'd prefer to widen it.

DevTechProfile:
CapFrameX test integration, see here: https://github.com/CXWorld/CapFrameX/blob/release/1.8.5/source/LibreHardwareMonitorLib/Hardware/Cpu/IntelCpu.cs

Core Ultra X7 358H / 0xCC
image

Core Ultra 9 285K / 0xC6
285K

namazso:
I think this PR may be superceded by https://github.com/namazso/PawnIO.Modules/pull/54

namazso:
> ### Point 1 — exposing the full 128 kB MCHBAR region
> This is the one that's worth a longer reply. The narrow surface is intentional, and the security envelope at the top of the file lists it as a design goal (IntelIMC.p:35-41):
> 
> > * Read-only (no writes to PCI cfg / MMIO / MSR / IO).
> > * No caller-controlled physical addresses or PCI BDF.
> > * Fixed compile-time MCHBAR offsets only.
> > * Strict CPUID allowlist; unknown models return `STATUS_NOT_SUPPORTED`.
> > * Reserved-bits and ratio-range checks reject wrong-register reads.
> > * MCHBAR enable bit is observed, never set.
> 
> The narrowness is what makes the safety checks meaningful:
> 
> * The reserved-bits mask in `read_memss_pma` and the `IMC_RATIO_MIN..MAX` range check work _because_ the module knows which register it's reading. A wrong-stepping or wrong-platform read fails closed.
> * A user-space tool can't accidentally turn an unknown MCHBAR field into a "Memory Clock" sensor.
> * The audit surface for a security review stays small and stable.
> 
> A generic `read_mchbar_dword(offset)` would lose all of that. MCHBAR isn't only IMC — it also covers DMI, PCIe, IMPH, etc. Reads are side-effect-free, agreed, but exposing arbitrary architectural reserved fields weakens the defense-in-depth story: the module would no longer be self-contained in what it reads.

None of this is a security issue though. A buggy tool would just get garbage data, but it can't bypass AV/EDR, bypass PPL, or open up a wider exploit primitive against the kernel. Simplicity is more important of a goal

namazso:
Superseded by #54

PR #54, opened 2026-05-03
Add Intel MCHBAR read module

a1ive:
This PR adds a read-only Intel MCHBAR module for PawnIO.

Supported Hardware : Intel Sandy Bridge (Core 2nd Gen) - Nova Lake

This module is read-only and does not expose any write operation to the MCHBAR region.
It rejects unsupported platforms, out-of-range offsets, and unaligned accesses.

Tested on:
- Alder Lake (i7-12700)
- Tiger Lake (Xeon W-11955M)

References:
- https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/intel-family.h
- https://github.com/memtest86plus/memtest86plus/tree/main/system/imc/x86
- https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/2nd-gen-core-desktop-vol-2-datasheet.pdf
- https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/4th-gen-core-family-desktop-vol-2-datasheet.pdf
- https://www.intel.com/content/www/us/en/content-details/631122/11th-generation-intel-coretm-processors-datasheet-volume-2-of-2.html
- https://www.intel.com/content/www/us/en/content-details/655259/12th-generation-intel-core-processor-datasheet-volume-2-of-2.html

### DIFF
--- a/IntelMCHBAR.p
+++ b/IntelMCHBAR.p
@@ -220,7 +220,7 @@ NTSTATUS:mchbar_init(CodeName:code_name) {
 DEFINE_IOCTL_SIZED(ioctl_read_dword, 1, 1) {
     new offset = in[0];
 
-    if (offset >= g_mchbar_size)
+    if (offset < 0 || offset >= g_mchbar_size)
         return STATUS_ACCESS_DENIED;
     if (offset & 0x3)
         return STATUS_ACCESS_DENIED;
@@ -242,7 +242,7 @@ DEFINE_IOCTL_SIZED(ioctl_read_dword, 1, 1) {
 DEFINE_IOCTL_SIZED(ioctl_read_qword, 1, 1) {
     new offset = in[0];
 
-    if (offset >= g_mchbar_size)
+    if (offset < 0 || offset >= g_mchbar_size)
         return STATUS_ACCESS_DENIED;
     if (offset & 0x7)
         return STATUS_ACCESS_DENIED;
--- a/IntelMCHBAR.p
+++ b/IntelMCHBAR.p
@@ -33,6 +33,170 @@
 
 new g_mchbar_addr = 0;
 
+// https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/intel-family.h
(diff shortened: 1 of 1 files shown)


namazso:
Looks fine, just one thing: `io_space_map` is actually quite expensive as synchronizing kernel address space requires an IPI and TLB flush. It'd be better to map it once and use the `unload` forward (see `Echo.p` for example) to unmap it on unload.

a1ive:
I noticed that `RyzenSMU.p` also seems to call `io_space_map` for each access, so I thought this usage pattern might be acceptable.

I also have another concern about mapping the whole MCHBAR region once. On Intel Comet Lake and earlier CPUs, the MCHBAR size is 64 KB or smaller, not 128 KB. If 128 KB is mapped unconditionally, it may map beyond the valid MCHBAR range on those platforms.

Would it be better to determine the actual MCHBAR size based on the CPU family/model, map only that range once, and then unmap it during unload?

namazso:
> I noticed that `RyzenSMU.p` also seems to call `io_space_map` for each access, so I thought this usage pattern might be acceptable.

That one should be updated at some point as well. The oldest PawnIO release had a bug where `unload` wouldn't get called, but that's quite irrelevant by now. It also does the whole page read in one go, so it's less likely to be spam-called by users.

> Would it be better to determine the actual MCHBAR size based on the CPU family/model, map only that range once, and then unmap it during unload?

Yes

a1ive:
I have updated the PR according to your suggestion.
The module now maps the MCHBAR region once during initialization and unmaps it in the unload forward.

The MCHBAR size is selected based on the CPU family/model, instead of always mapping 128 KB:

32 KB for pre-IceLake microarchitectures
64 KB for IceLake through RocketLake
128 KB for TigerLake and newer

I also removed the server and some special-platform model entries for now, because I could not find public documentation for their MCHBAR register layout.

Tested on:

- Alder Lake (i7-12700)
- Tiger Lake (Xeon W-11955M)

PR #56, opened 2026-05-11
AMD RyzenSmu updates and fixes

irusanov:
- Adds a missing command for Zen and Zen+, without it the table is not transferred to RAM
- Map io space once, unmap on unload
- Reuse max args constant instead of hardcoded 6 number of arguments


irusanov:
@namazso Just realized how dumb I am. I'm now calling ioctl (to resolve pm table) in the main method, so a pci lock is needed when loading the module and that is not desired. Moreover not all clients might need the pm table.

I will fix it today by moving the map to the respective ioctl, but will only map it once by checking the global map pointer. Sorry for the inconvenince this might have caused to other clients using this module.

PR #68, opened 2026-06-23
Fixes a SMBus controller hang after a block read

Blacktempel:
Fixes a SMBus controller hang after a block read, which occured when testing with DDR3 RAM.

LAST_BYTE is now set in time before the final byte so that the controller knows when the transfer ends.
Previously, LAST_BYTE was set too late, causing the controller to hang and subsequent SMBus accesses to fail.
Additionally, any pending BYTE_DONE is acknowledged before KILL so that aborted transfers can be cleanly terminated.

### DIFF
--- a/SmbusI801.p
+++ b/SmbusI801.p
@@ -303,13 +303,20 @@ NTSTATUS:i801_check_pre()
 }
 
 i801_kill() {
+    // In byte-by-byte block mode the controller keeps SMBCLK low while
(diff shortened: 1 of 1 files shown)


Blacktempel:
@namazso is it possible to merge this week please ? Thanks.

Blacktempel:
@namazso could you create a new release aswell, please ?

PR #70, opened 2026-07-01
Add AppleSMC module: Apple SMC fan/sensor access

JdovGamester:
## Summary
- Adds `AppleSMC.p`, a module for reading/writing Apple SMC keys (fan RPM, min/max, target, mode bitmask, temperature sensors, etc.) on Intel Macs running Windows via Boot Camp.
- Talks to the SMC over its standard I/O-port protocol (data port `0x300`, command port `0x304`) — the same interface documented and independently implemented by e.g. Linux's `drivers/hwmon/applesmc.c`.
- Deliberately scoped to **only** those two fixed ports (unlike `LpcIO.p`'s BAR-discovered range) — this module can't be used to poke arbitrary I/O, only the well-known Apple SMC handshake.
- Exposes two IOCTLs: `ioctl_smc_read(key, len) -> data` and `ioctl_smc_write(key, len, data)`.

## Motivation
Built for a personal Boot Camp power/thermal management project. Wanted real fan-speed control without WinRing0 (blocklisted, CVE-2020-14979) — PawnIO's signed, scoped-module model is exactly the right fit, but there's currently no Apple SMC module in this repo.

## Test plan
- Compiles cleanly with `pawncc -C64` against this repo's `include/` (only pre-existing "tag mismatch" style warnings, same category as other modules comparing `NTSTATUS:` return values).
- Verified with `PawnIOUtil test` against the signed driver: module load correctly fails at the signature-trust step (expected, since it isn't signed by the maintainer key) — confirms the module and PawnIOLib integration are structurally sound up to that boundary.
- Full read/write behavior (SMC handshake, fan RPM decode/encode, `FS!` mode bitmask) is implemented per the well-documented protocol but untestable end-to-end until signed, since I don't have a way to load it through the trusted driver myself.

Happy to adjust naming/style/scope to match repo conventions — first contribution here.

### DIFF
--- a/AppleSMC.p
+++ b/AppleSMC.p
@@ -1,0 +1,181 @@
+//  PawnIO Modules - Modules for various hardware to be used with PawnIO.
+//  Copyright (C) 2026  (module author to be filled in on PR)
+//
+//  This library is free software; you can redistribute it and/or
+//  modify it under the terms of the GNU Lesser General Public
+//  License as published by the Free Software Foundation; either
+//  version 2.1 of the License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+//  Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public
+//  License along with this library; if not, write to the Free Software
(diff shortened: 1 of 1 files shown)


namazso:
Was this fully tested with the unrestricted driver? Note that the 2.2.0 installer is bugged, you'll need 2.1.0 or build the unrestricted version yourself.

Additionally, is it possible to restrict it more in `main` to only Apple devices at least? ~~I'd assume there's some easily accessible artifact of an Apple device, unlike the other modules where manufacturers just throw random chips together on the board.~~ You can see how DMI checks are done in ValveLeds.p

PR #77, opened 2026-07-14
Add NVIDIA GPU PawnIO module

DevTechProfile:
This PR adds a new dedicated PawnIO module for direct NVIDIA GPU hardware readings. The initial version provides hotspot data for GB20x GPUs and can be extended with more readings later.

Tested on RTX 5090.

image


Sakura-Lhy0409:
Independent corroboration from HWMonitor 1.65 x64 reverse engineering and live testing on an NVIDIA GeForce RTX 5090 Laptop GPU (PCI 02:00.0, device ID 10DE:2C18):

- HWMonitor's NVIDIA thermal provider is at `HWMonitor_x64.exe + 0xA88B0`.
- It gates this path with `(gpuClass & 0xFFFFF000) == 0x010D0000` before any MMIO access.
- Sensor ID 2 (`Hot Spot`) reads BAR0 offset `0xAD0A90`.
- Sensor ID 34 (`HotSpot #2`) reads BAR0 offset `0xAD0AE0`.
- The raw value is decoded as `(raw >> 8 & 0xFF) + (raw & 0xFF) * 0.03125 C`; integer bytes `0x00` and `0xFF` are rejected as invalid.
- Live readings were 47 C and 48 C while HWMonitor displayed the same two values.
- On the same GPU, `NvAPI_GPU_GetThermalSettings` exposes only the core GPU sensor, so these two readings are not available through the ordinary user-mode thermal API.

The fixed offsets, NVIDIA display-class check, BAR validation, GB20x `NV_PMC_BOOT_0` allowlist, and read-only IOCTL in this PR match the observed implementation and keep the module narrowly scoped. This would let M-Control expose the two hotspot sensors as optional additional telemetry without replacing its existing GPUMON core/memory temperatures.

DevTechProfile:
Thanks for taking the time to reverse-engineer this and independently verify it on an RTX 5090 Laptop GPU. It’s
great to have confirmation that both MMIO offsets and the raw value format match HWMonitor’s implementation.

The PawnIO module intentionally returns the raw register values. In CapFrameX, the NvidiaThermal wrapper applies the
same decoding formula and treats an integer byte of 0x00 or 0xFF as invalid, so that part is already handled on the
consumer side. I think it's a good approach to have as little logic in the module as possible.

It’s interesting that the module could be useful for M-Control.

Blacktempel:
I forgot to comment:
the read works for me too - RTX 5090.

DevTechProfile:
Great, thanks for the confirmation.

namazso:
I think this is the first removable device anyone ever tried to support. I guess since we check target device each request there isn't a long enough timespan for issues stemming from that.

Nvidia.p
```diff
@@ -0,0 +1,160 @@
...
+    new bar0 = 0;
+    new NTSTATUS:status = get_bar0(bus, device, function, bar0);
+    if (!NT_SUCCESS(status))
+        return status;
+
+    status = check_gpu_architecture(bar0);
+    if (!NT_SUCCESS(status))
+        return status;
+
+    new VA:va = io_space_map(bar0 + THERMAL_MMIO_BASE, THERMAL_MMIO_SIZE);
```
namazso: As discussed in https://github.com/namazso/PawnIO.Modules/pull/54#issuecomment-4366395090 mapping changes page tables and requires an IPI, which is relatively expensive and probably not very good when you're trying to do this during gameplay. Could you keep one mapping around, and only remap if we end up with a different address? You can unmap it in unload. This should at least benefit one GPU machines.

namazso (changes requested):

GDuqueB:
Hi, I independently opened #78 before noticing this PR. Both implementations use the same two GB20x thermal registers.

Regarding the mapping overhead mentioned in the review, #78 maps a single 4 KiB BAR0 page during initialization, keeps it mapped for subsequent reads, and releases it on module unload. That pattern may be useful here to avoid changing page tables on every sample.

#78 is deliberately limited to the tested RTX 5090 PCI ID `10DE:2B85`, whereas this PR has broader GB20x architecture validation and more runtime testing. For removable-device safety, a cached implementation should still revalidate the target BDF/device before each read and remap if the BAR changes or the device is replaced.

I would be happy to collaborate and consolidate the useful parts of both implementations. Please let me know which PR you would prefer to use as the base. My goal is to obtain a narrowly scoped signed module that can be consumed by a FanControl plugin.

Nvidia.p
DevTechProfile: Yeah, good point. The current hot path actually maps twice per request, once for NV_PMC_BOOT_0 and once for the thermal registers. I’ll just keep the per-request PCI/BAR validation, cache the thermal mapping keyed by BDF/device identity/BAR0, rerun the architecture check only when that key really changes, and unmap everything in unload. If the device disappears or BAR0 changes, the cached VA will not be accessed. I hope this works out.

Retest on RTX 5090

image

DevTechProfile:
Thanks, this aligns with the update I’ve just pushed. PR #77 now keeps the thermal MMIO mapping cached. The GB20x architecture check is repeated whenever the cache target changes.

Since #77 already has broader GB20x validation and additional runtime testing, I think it makes sense to keep #77 as
the base, but I’m happy to consolidate any remaining useful details from #78.

GDuqueB:
Thanks, using #77 as the base makes sense to me. I have reviewed the updated mapping approach and I do not see any remaining module-level detail from #78 that needs to be transferred at this point.

Once a signed test artifact is available, I would be happy to validate it on my RTX 5090 Founders Edition. My intention is to develop a free, non-commercial FanControl plugin for personal and community use. It would expose the hotspot readings through PawnIO and obtain the GPU core and memory-junction temperatures through NVAPI, removing the need to keep GPU-Z running.

Please let me know when a signed test build is available.

c-michail:
Can you do the same with AMD/Intel GPUs ?

DevTechProfile:
Yes, I have plans at least for AMD cards. Some cache metrics for Nv cards would be nice as well.

iamnotstanley:
Hi,

Sorry for jumping in uninvited. I really appreciate your work and hope this improvement can be used in projects like FanControl and LibreHardwareMonitor soon.

I have two questions about this PR:

HWMonitor just released version 1.65.1 with a fix for Blackwell hotspot readings. I’m not familiar with reverse engineering, but could their fix be useful to look at?

Also, I followed Unwinder's comments about this topic and he mentioned that six thermal channels can be read from the MMIO registers ([post 1](https://www.overclock.net/threads/official-nvidia-rtx-5090-owners-club.1814246/page-1907?post_id=29600182#post-29600182), [post 2](https://forums.guru3d.com/threads/msi-afterburner-v4-6-7-beta-1-is-approaching.458132/post-6446669)). Would it be possible to expose all six, or perhaps add a third value showing the maximum?

Thanks!

GDuqueB:
Important correction from HWMonitor 1.65.1: the original 1.65 register decoding is not valid.

I compared the x64 routines in HWMonitor 1.65 and 1.65.1 (file version `1.6.5.1`, SHA-256 `C9C3B20F8B8CFBDCDDE852E03CBD42934E7E946660AC7877057F3143059211D3`). The corrected implementation does the following:

- keeps BAR0 offset `0xAD0A90` for the displayed Hot Spot value;
- no longer references `0xAD0AE0` (`HotSpot #2` in 1.65);
- reads six contiguous channels at `0xAD0A90`, `0xAD0A94`, `0xAD0A98`, `0xAD0A9C`, `0xAD0AA0`, and `0xAD0AA4`;
- tests bit 30 and rejects the sample when that bit is clear;
- converts a valid sample as `(raw & 0xFFFF) * 0.00390625`, i.e. `(raw & 0xFFFF) / 256.0` degrees Celsius.

Therefore the old byte-based formula `((raw >> 8) & 0xFF) + (raw & 0xFF) / 32.0` and the `0xAD0AE0` second value should not be used by consumers. The PawnIO module can still return raw values, but its read contract should expose either the corrected main register or all six contiguous DWORDs; consumers such as CapFrameX/FanControl need the bit-30 validity check and `/256.0` scale.

I have updated #78 in commit `64a783d` to expose the six raw channels and document the corrected consumer contract. I have not yet performed a signed live read, so runtime comparison against HWMonitor 1.65.1 remains desirable.

DevTechProfile:
Ok thanks, will check that later. I would prefer the module exposing the raw values, but I will add a comment for consumers.

namazso:
I’d recommend raw values where possible, since PawnIO has no float support (and is not the place to do this anyway, it’s meant to be just the bare minimum needed for access control)

Blacktempel:
Works fine (not the code from this PR, but tested all sensors). Tested on RTX 5060 Ti aswell.

grafik

HWMonitor displays same values.

DevTechProfile:
> I’d recommend raw values where possible, since PawnIO has no float support (and is not the place to do this anyway, it’s meant to be just the bare minimum needed for access control)

Done.  The module now reads the six contiguous DWORDs at BAR0 offsets 0xAD0A90 through 0xAD0AA4. The obsolete 0xAD0AE0 access has been removed.

DevTechProfile:
I have a question @namazso: What’s the better approach, a general GPU module, or separate context-specific modules for thermals, power, fan, and so on?

Liddo-kun:
I'm preparing hardware validation of this module on an **NVIDIA RTX 5090 Laptop GPU** (`10DE:2C18`) 

Done so far against head `e1b8f13`:
- Compiled `Nvidia.p` with Pawn 4.1.7152 and the exact CI flags, no warnings
- Consumer integration (app + standalone capture harness) doing all decoding in userspace per the raw-values approach: bit-30 validity, `(raw & 0xFFFF) / 256.0`, plausibility bounds, BDF supplied from passive SetupAPI discovery.
- Planned matrix: idle / load / cooldown vs HWMonitor 1.65.1, `0xAD0AA0`-vs-individual-channel relationship, power-gating (all-ones) behavior, sleep/resume BAR revalidation.

One source observation from the laptop angle: when the dGPU is deeply power-gated (D3cold), the PCI config-space reads in `get_bar0` return FFs, so the vendor check fails and the IOCTL returns `STATUS_NOT_SUPPORTED` — rather than the all-ones data pattern. Consumers should treat `NOT_SUPPORTED` after previously-successful reads as "likely asleep", not "unsupported". Might be worth a doc comment on the IOCTL; no code change needed as far as I can see.

Currently blocked on actually loading the module: the public 2.2.0 unrestricted edition still enforces module signatures (namazso/PawnIO.Setup#11), so an unsigned dev build can't load. @namazso — would you consider a signed test build of `Nvidia.bin` at the current head (or including it in a modules release once review is settled)? I'll run the full matrix on the laptop GPU and post results here.

Blacktempel:
Use an older version of PawnIO, that works with current modules aswell.

namazso:
I’ll try to get this reviewed at the latest this weekend, been busy recently

DevTechProfile:
> I’ll try to get this reviewed at the latest this weekend, been busy recently

No problem. It would be great to have the module ready for CapFrameX v1.9.0.

To be honest, I still have a lot to test and fix before the release. CapFrameX is getting its own OSD, which has turned out to be a much bigger project than I initially expected. On top of that, my AMD card has been crashing all the time lately, which certainly isn't helping.

At this rate, I guess you'll be faster than me... ^^

Liddo-kun:
Update: priorities shifted on my end and I won't be pursuing this validation after all — please don't go out of your way to produce a signed test build on my account. Apologies for the noise.

The D3cold observation above still stands for whoever picks up laptop testing: a deeply power-gated mobile dGPU fails `get_bar0`'s config-space reads (`STATUS_NOT_SUPPORTED`) rather than returning all-ones data, so consumers should treat that as likely-asleep after previously-successful reads. Good luck with the PR.

PR #78, opened 2026-07-16
Add read-only NVIDIA GB202 thermal channels

GDuqueB:
## Summary

This PR adds a narrowly scoped PawnIO module for reading the six contiguous GB202 thermal channels used by HWMonitor 1.65.1 on the NVIDIA GeForce RTX 5090 (PCI ID `10DE:2B85`).

The module:

- accepts a caller-supplied PCI BDF and verifies the exact vendor/device ID;
- resolves BAR0 from PCI configuration space;
- maps only the 4 KiB page at BAR0-relative offset `0x00AD0000`;
- exposes six fixed read-only DWORDs from `0x00AD0A90` through `0x00AD0AA4`;
- provides no MMIO, PCI configuration, I/O-port, MSR, or physical-memory write primitive;
- supports x64 only and releases the mapping on unload.

## HWMonitor 1.65.1 correction

HWMonitor 1.65.1, released after the original submission, corrected its RTX 50 hotspot implementation. A static comparison with 1.65 shows that:

- `0x00AD0A90` remains the register used for the displayed Hot Spot value;
- `0x00AD0AE0`, previously interpreted as `HotSpot #2`, is no longer read;
- six channels are now read at `0x00AD0A90`, `0x00AD0A94`, `0x00AD0A98`, `0x00AD0A9C`, `0x00AD0AA0`, and `0x00AD0AA4`;
- bit 30 must be set for a sample to be valid;
- a valid sample is decoded as `temperature_c = (raw & 0xFFFF) / 256.0`.

The module intentionally returns the six raw DWORDs. Validation and temperature conversion remain the responsibility of the userspace consumer.

## Motivation

Recent NVIDIA drivers no longer expose a reliable RTX 50 hotspot value through the interface previously used by monitoring plugins. HWMonitor 1.65.1 obtains the value through these GB202 BAR0 registers rather than through an ordinary NVAPI thermal call.

The intended consumer is a free FanControl plugin. GPU core and GDDR7 memory-junction temperatures have been independently validated through `NvAPI_GPU_GetThermalSensors` on the same RTX 5090 and will remain in userspace; PawnIO is needed only for the Blackwell thermal channels unavailable through that interface.

## Safety constraints

The module is deliberately device-specific and read-only. It rejects every PCI device other than `10DE:2B85` and does not expose arbitrary offsets or addresses to userspace. All six DWORD offsets are constants in the signed module.

## Validation performed

- Compiles with Pawn compiler 4.1.7152 using the repository CI flags, with no warnings.
- Source review confirms that no write primitive is used.
- Static comparison of HWMonitor 1.65 and 1.65.1 confirms the corrected register set, validity test, and scale factor.
- The target machine identifies the GPU as `10DE:2B85` at `0000:01:00.0`.
- The unsigned module is rejected by official PawnIO 2.2.0 as expected.

An actual register read with this exact module remains pending because the test system has Secure Boot enabled and official PawnIO accepts only signed modules. The owner of the tested RTX 5090 Founders Edition is available to compare a signed test build with HWMonitor 1.65.1 at idle and under load.

## Scope

This PR contains only `NvidiaGB202Thermal.p`. It does not include proprietary CPUID binaries, disassembly output, memory dumps, logs, or experimental userspace probes.

### DIFF
--- a/NvidiaGB202Thermal.p
+++ b/NvidiaGB202Thermal.p
@@ -1,4 +1,4 @@
-// PawnIO module for NVIDIA GB202 thermal registers observed in HWMonitor 1.65.
+// PawnIO module for NVIDIA GB202 thermal registers observed in HWMonitor 1.65.1.
 // Copyright (C) 2026 Gonzalo Duque de Blas (@GDuqueB)
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
@@ -14,11 +14,11 @@
 #define PCI_BAR_MEM_TYPE_64BIT     0x04
 #define PCI_BAR_MEM_ADDR_MASK      0xFFFFFFF0
 
-// Both registers fit in this single 4 KiB page of BAR0.
+// All six thermal channels fit in this single 4 KiB page of BAR0.
 #define THERMAL_PAGE_OFFSET        0x00AD0000
 #define THERMAL_PAGE_SIZE          0x1000
-#define HOTSPOT_PAGE_OFFSET        0x0A90
-#define HOTSPOT_2_PAGE_OFFSET      0x0AE0
+#define THERMAL_CHANNELS_OFFSET    0x0A90
+#define THERMAL_CHANNEL_COUNT      6
 
 new VA:g_thermal_page_va = NULL;
 new g_pci_bdf = 0;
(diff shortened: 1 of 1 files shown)


namazso:
I think this was replaced by https://github.com/namazso/PawnIO.Modules/pull/77

If not, please ping me

PR #79, opened 2026-07-18
Add read-only Intel PCH thermal module

FengXi7420:
## Summary

- add a read-only Intel PCH thermal module
- support the documented PCI thermal BAR0 layout used by `intel_pch_thermal`
- support the fixed PWRM `0xFE000000 + 0x1560` layout used by recent Intel client PCH generations
- expose only raw temperature-register and identity reads; no PCI/MMIO write native is referenced

## Consumer contract

The module deliberately returns raw values and keeps temperature semantics in the consumer:

- PCI BAR0 route: validate `raw & 0x1ff` against `0` / `0x1ff`, then convert with `(raw & 0x1ff) * 0.5 - 50 C`
- fixed PWRM route: require bit 9, validate `raw & 0x1ff` against `0` / `0x1ff`, then use `raw & 0x1ff` as integer Celsius
- `ioctl_read_raw` returns `{ raw, route }`
- `ioctl_identity` returns the selected route, PCI identity/BDF, and mapped MMIO range

The modern route is gated by the LPC/eSPI device IDs observed in HWMonitor 1.65 before mapping the fixed window. Initialization validates only device identity and mapping; transient raw-sample validity remains a consumer concern.

## Validation

- compiled all 21 modules successfully with Pawn 4.1.7152 and the repository CI flags: `-iinclude -C64 -;+ -(+ -p`
- verified the module source contains no `physical_write_*`, `pci_config_write_*`, `virtual_write_*`, I/O write, or MSR write primitive
- validated the consumer-side conversion and invalid-sample vectors for both routes
- tested the modern route on an Intel HM870 platform (`PCI VEN_8086 DEV_7F0C`): a read-only physical-memory probe of `0xFE001560` returned `0x80000231`; bit 9 was set and the consumer-side decoded temperature was 49 C
- the MSI EC PCH proxy (`EC 0x9E`) read 43 C during the same validation session, consistent with a separate firmware sampling path

The stock PawnIO driver accepts signed modules only, so final loading of this new module requires a maintainer-signed build from the upstream release workflow.

## Sources

- Linux `drivers/thermal/intel/intel_pch_thermal.c` for the PCI BAR0 register, device IDs, mask, resolution, and offset
- static and runtime analysis of HWMonitor 1.65 for the recent-PCH fixed-window route, LPC/eSPI allow-list, valid bit, and integer-Celsius consumer contract

### DIFF
--- a/IntelPCHThermal.p
+++ b/IntelPCHThermal.p
@@ -1,0 +1,302 @@
+//  PawnIO Modules - Modules for various hardware to be used with PawnIO.
+//  Copyright (C) 2026  M-Control Contributors
+//
+//  This library is free software; you can redistribute it and/or
+//  modify it under the terms of the GNU Lesser General Public
+//  License as published by the Free Software Foundation; either
+//  version 2.1 of the License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+//  Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public
+//  License along with this library; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+//  SPDX-License-Identifier: LGPL-2.1-or-later
+
+#include <pawnio.inc>
+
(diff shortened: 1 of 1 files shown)


FengXi7420:
Updated the module to follow the raw-value/minimal-access-control design discussed in #77: `ioctl_read_raw` now returns only `{ raw, route }`, initialization no longer interprets sample validity, and all validity checks/conversion are documented as consumer responsibilities. Recompiled successfully with Pawn 4.1.7152.

PR #80, opened 2026-07-18
RyzenSMU: Add Sensor monitoring support for pre-Ryzen CPUs, add more supported pre-Ryzen CPU codenames

Erruar:
Tested on real hardware. Minimal changes to code. Shouldn't affect on Ryzen CPUs


irusanov:
This changes the CodeName and shifts existing indicies by adding new codenames in the beginning. Current list is not chronologically ordered anyway, so for keeping it simpler and in case other apps rely on the enum (via ioctl_get_code_name), I think it might better to add any new codenames at the end, regardless of generation.

Erruar:
cmds
Completely tested on Bristol Ridge, and soon will be tested on Stoney Ridge

Erruar:
Tested on Stoney Ridge comfirmed working. @namazso could you please accept my changes?

namazso:
Not a huge fan of how it had to be special cased throughout the whole module, but it does share at least *some* code with the rest.

Issue #86, opened 2026-07-28
PawnIO Module for controlling Valve Steam Machine LEDs


CalcProgrammer1:
Opening an issue to track progress, as I plan to implement this myself.  I got my Steam Machine last week and added support for it in OpenRGB under Linux based on Valve's own leds-valve.c driver, which exposes controls via sysfs (/sys/class/leds/valve-led[0-16]).  I made a DKMS package with this driver so it can be installed on other distros.  The interface is a simple regmap starting at IO port address 0x0DE8.  There are register addresses for the various LED colors, brightness settings, enable settings, persistence settings, delay/speed, and mode selection.  I just poked at it with RWEverything and was able to control the LEDs pretty easily just reading the register definitions at the top of leds-valve.c.  It should be fairly straightforward to convert to a PawnIO module.

<img width="1680" height="1050" alt="Image" src="https://github.com/user-attachments/assets/e705c13b-dc94-4a47-925e-e469cd363519" />

Reference: https://gitlab.com/CalcProgrammer1/leds-valve-dkms/-/blob/master/src/leds-valve.c

Edit: Making good progress already, my basically 1 to 1 PawnIO module based off of leds-valve can set the LEDs and mode

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/5c7089a7-5023-45dd-af4a-52bea51bf051" />

CalcProgrammer1:
@namazso are there functions within PawnIO for getting/verifying DMI data?  Valve's official driver uses DMI string filters to only activate the driver on the intended hardware.  I can do this outside of the module in OpenRGB and only load in the module if the DMI info matches, but this would not prevent other applications from erroneously loading in the module on an unintended system.

namazso:
Unfortunately Windows doesn't have such functionality (that is easily accessible) so PawnIO doesn't have it either. We could potentially use the registry: `\Registry\Machine\HARDWARE\DESCRIPTION\System\BIOS` but that needs some abstraction, probably on the lengthier side of things. The only other alternative Windows provides is ExGetSystemFirmwareTable and parsing the SMBIOS tables on your own, probably not a very fun task.

CalcProgrammer1:
Implemented in https://github.com/namazso/PawnIO.Modules/pull/87

PR #87, opened 2026-07-29
Initial commit of LedsValve.p, LED driver for Valve Steam Machine

CalcProgrammer1:
Still needs some work, but this was a quick port from the leds-valve.c Linux driver.  It compiles and I have successfully used it to control the Valve Steam Machine with OpenRGB to the same level of functionality my Linux implementation has (direct control over independent LEDs, selecting built in effects, adjusting speed of built in effects, and adjusting brightness).  Does not perform DMI verification of the vendor/product like the leds-valve.c driver does as PawnIO does not support DMI, so I'm doing that verification on the OpenRGB side before loading this module.  Not all ioctls have been tested yet.

Implements #86 

OpenRGB side changes: https://gitlab.com/CalcProgrammer1/OpenRGB/-/merge_requests/3439


CalcProgrammer1:
I've done some code cleanup and now I think it's ready for review.

One thing I'm still deciding on is whether to pass the RGB values in 3 separate Pawn cells or combine them into a single cell, it might be more efficient to do single cell and if I match the 0x00BBGGRR format that simplifies the OpenRGB side as that is the format our RGBColor type uses.

I renamed it from leds-valve.p to LedsValve.p to match the other module naming conventions.  The ioctls implemented line up with the leds-valve.c sysfs entries in /sys/class/leds/valve-led[0-16] on Linux with the addition of ioctl_init for initializing the module.

CalcProgrammer1:
Another thing I am thinking of adding is an ioctl_get_led_count or something along those lines, as in the Linux implementation you can count the number of /sys/class/leds/valve-led[X] entries to determine LED count but here this method is not available.

@namazso have any thoughts on this MR?  Ideally these ioctls, at least the ones from the common Linux framework, could be reused on other Linux multi-LED/RGB driver ports in the future.

namazso:
I just pushed a new "registry" branch, could you try using it to check baseboard vendor? `\Registry\Machine\HARDWARE\DESCRIPTION\System\BIOS` should contain various keys from SMBIOS.

CalcProgrammer1:
I'll look into testing the registry check this week.

CalcProgrammer1:
Implemented the registry check and verified on the Steam Machine.  I moved the registry.inc file into the includes folder to get it to compile properly.  I also had to write a string comparison function locally to compare the strings, that might be a good candidate to move to an include file (unless there's already a string compare function I am missing that I could use instead).

CalcProgrammer1:
Updated the merge request to add the registry check within the PawnIO module using your new registry include file.  Confirmed that it works and intentionally corrupted the strings it checks the two parameters against to force it to fail and confirm it does not load even if OpenRGB attempts to load it.

CalcProgrammer1:
Ready for review/merge.  I cleaned up the registry info check (the original check I had would allow incorrect pairings OEM:Fremont and Valve:F7F where the Linux driver would not).  I added functions for getting the number of LEDS and number of components for each LED, which could be useful if we reuse the same ioctls for future Linux-ported multi_intensity LED drivers.  Verified latest build on hardware including the new ioctls.

CalcProgrammer1:
Removed all internal state tracking/global variables.  Replaced with register reads where appropriate.

PR #93, opened 2026-08-19
Add GigabyteID.p Module

TsunamiMommy:
Add's GigabyteID.p module used to get two specific motherboard identification numbers used for determining feature support and layout information for fan, voltage, and temperature monitoring and RGB lighting control. Let me know if there are any issues you see with it. It's pretty barebones.


namazso (approved):
Thanks for getting this tested, this will need the PawnIO 2.3.0 release (whenever that happens)

PR #95, opened 2026-08-23
Add read-only Acer A315-56 extended EC module

celsonaemen:
## Summary

Adds a model-specific, read-only semantic backend for the extended embedded-controller mailbox exposed by the Acer Aspire A315-56 ACPI firmware.

The module exposes one IOCTL that implements the semantic equivalent of EC0.FANG(index). Although the caller operation is a read, the firmware protocol necessarily writes the index and command fields before returning ERN1.

## Hardware and firmware gate

Validated target:

- Acer Aspire A315-56
- mainboard Sleepy_IL
- Insyde BIOS V1.27
- Intel Core i3-1005G1, family 6/model 0x7E
- Intel LPC/eSPI PCI ID 8086:3482

The module refuses to load unless the CPU architecture/vendor/family/model and exact LPC vendor/device ID match.

## Firmware reference

The transaction was derived from the decompiled DSDT path EC0.FANG -> EIDR/ERRD -> ERCD:

- ERIB: 0xFE0B0800 + 0x05D
- ECTB/ECTE: 0xFE0B0800 + 0x582
- ERN1: 0xFE0B0800 + 0x583
- ECMD: 0xFE0B0800 + 0x58B
- EDT1/EDT2: 0xFE0B0800 + 0x58C/0x58D
- extended-read command: ECMD 0xB0, EDT1 0x5F, EDT2 0

A sanitized hardware map and the reverse-engineering notes are available at:
https://github.com/celsonaemen/acer-a315-56-control/blob/main/docs/HARDWARE-MAP.md

## Safety properties

- no arbitrary MMIO, EC, or fan-value write IOCTL
- one fixed 4 KiB MMIO mapping
- rejects an out-of-range 16-bit index
- returns STATUS_DEVICE_BUSY when ECTB or ECTE is already set
- documents the Access_EC user-mode mutant requirement
- one-second bounded poll with STATUS_IO_TIMEOUT
- best-effort ECTB cleanup on every post-acquisition exit
- LGPL-2.1-or-later with Doxygen documentation

## Validation

- compiled with bundled Pawn 4.1.7152 and the repository CI flags
- all 23 top-level modules compiled successfully with no warnings
- git diff --check passed
- the unsigned module has not been loaded on the restricted production driver

The main review request is the ACPI/Acpi.sys concurrency model around the firmware mailbox and whether the hardware gate is sufficiently specific for an official signed build.

### DIFF
--- a/AcerA31556EC.p
+++ b/AcerA31556EC.p
@@ -1,0 +1,175 @@
+//  PawnIO Modules - Modules for various hardware to be used with PawnIO.
+//  Copyright (C) 2026  Celso Naemen <180766906+celsonaemen@users.noreply.github.com>
+//
+//  This library is free software; you can redistribute it and/or
+//  modify it under the terms of the GNU Lesser General Public
+//  License as published by the Free Software Foundation; either
+//  version 2.1 of the License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+//  Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public
+//  License along with this library; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+//  SPDX-License-Identifier: LGPL-2.1-or-later
+
+#include <pawnio.inc>
+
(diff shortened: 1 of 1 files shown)


namazso:
I changed this to draft, please test it before marking it as ready.

celsonaemen:
A causa ficou fechada: dentro do CAB há quatro slots esperados (ARM64 normal/unrestricted e x64 normal/unrestricted), mas o quarto slot repete
  exatamente o ARM64 unrestricted. O x64 normal existe; o x64 unrestricted não foi empacotado.
  
  but i gona test it keep calm pls

PR #104, opened 2026-08-25
Add IntelRaptorLakeOcMailbox: narrow MSR 0x150 OC-mailbox module for Raptor Lake-S

ObnubiladO:
﻿This adds a narrow PawnIO module for the Intel OC-mailbox (MSR 0x150) used by Raptor Lake-S undervolting. It exposes exactly four typed operations — query IA/Core global offset, query Ring/Cache global offset, set non-positive IA/Core offset, set non-positive Ring/Cache offset — with **no caller-selected MSR, domain, command, or raw 64-bit request**.

## Why a dedicated module instead of IntelMSR

PawnIO.Modules already exposes OC-mailbox writes through the official IntelMSR module. IntelMSR provides a broader allowlisted MSR-write interface (0x150 OC mailbox AND package power limits AND VR-configuration mailbox AND hardware-prefetch/cache control). This module intentionally **reduces the privileged surface** relative to using IntelMSR:

| | IntelMSR | IntelRaptorLakeOcMailbox |
|---|---|---|
| MSR-write interface | broader allowlist | MSR 0x150 only |
| CPU scope | any Intel | family 6 / model 0xB7 / stepping 1 only |
| Operations | caller-selected MSR + value | fixed query/set operations |
| Domains | caller-chosen | fixed (0 IA/Core, 2 Ring/Cache) |
| Caller-selected MSR | yes | no |
| Caller-selected command | yes | no |
| Raw mailbox packet | possible | no |
| Positive offset | possible | no (non-positive only) |
| Power/VR/ratio writes | possible | no |

An application that only needs the OC mailbox should not hold a module capable of other privileged writes. This reduction is the primary reason for a separate narrow module.

## Design

- MSR `0x150` only; the internal mailbox helper rejects any other address.
- CPU gate: family 6, model 0xB7, stepping 1 (the tested i5-13600KF). The module's `main()` returns `STATUS_NOT_SUPPORTED` elsewhere, so it refuses to load on unverified CPUs (per the wiki guidance to "do best effort to ensure it only runs with supported hardware"; the module itself imposes no Pawn-side lockdown).
- SET is **non-positive offset only**: `raw_steps <= 0` — zero is accepted (an exact restore state may be 0 offset) and negatives are accepted; positive offsets and reserved bits are rejected before any privileged access. Payload is checked against the signed 11-bit representation (bits [31:21], 1/1024 V step).
- Query and SET packets (0x10 / 0x11, domains 0/2) are line-confirmed against five independent implementations; see the evidence section below.
- Bounded busy polling (max 8 spins, 50 µs sleep, ~400 µs worst case); timeout fails; mutations are never retried silently.
- No OC-lock/UVP/CFG-lock bypass, no NVRAM, no firmware or security-state access of any kind. Intended for the normal restricted PawnIO driver (module signature verified against the embedded trusted-key list).

## Security context (explicit, per contribution guidelines)

Runtime OC-mailbox voltage control is **security-sensitive**: it changes CPU operating voltage, and undervolting has real security implications (it is the basis of fault-injection attacks such as Plundervolt). This module does **not** claim otherwise. It is offered so that a narrow, auditable, signed surface exists for a capability the platform already exposes.

- This module does **not** bypass Undervolt Protection (UVP), OC Lock, firmware policy, VBS/HVCI, Secure Boot, or any other security control.
- It operates only when the platform itself permits the mailbox operation; it **cannot enable a firmware-disabled capability**.
- No security-bypass operation exists in the module.
- The maintainer decides whether this security-sensitive capability meets upstream policy.

## Evidence / protocol references

The packet layout (MSR 0x150; cmd 0x10 query / 0x11 set; domain 0 IA/Core, 2 Ring/Cache; payload bits [31:21], 1/1024 V step) is line-confirmed against immutable commits of independent implementations:

- georgewhewell/undervolt `2d825f96` — `undervolt.py` `pack_offset()` / `convert_rounded_offset()` / readback `unpack_offset()`
- kitsunyan/intel-undervolt `ea0e74c5` — `undervolt.c` (~lines 52-63) rdval/wrval + readback compare
- wesmar/UnderVolter `716df96a` — `src/VfCurve.c` `IAPERF_ProbeDomainVF` / `IAPERF_ProgramDomainVF`, `src/OcMailbox.c`, `src/CpuData.c` (`{6,183,1}` "RaptorLake" RPL-S B0), `src/Platform.h` `VoltDomains`
- subratamal/VoltageShift `33aab6ae` and xCuri0/VoltageShiftSecure `6b4612c0` — `main.mm` `writeOCMailBox` / `readOCMailBox`

Tested target: Intel Core i5-13600KF, family 6, model 0xB7 (183), stepping 1 (CPUID `0x000B0671`). See also Intel's Undervolt Protection documentation (Intel support article 000094219).

## Build

Compiles with the pinned pawncc 4.1.7152, zero warnings, with the repository's `-C64` flags. Compiled AMX bytecode is audited offline (only the four typed IOCTLs exported; only the allowed natives imported; every `msr_read`/`msr_write` call is inside the 0x150-guarded helper).

## Known limitations

Only the tested Raptor Lake-S B0 stepping is gated; absolute-vs-additive mailbox semantics on this platform are unverified and require a runtime readback study; no claim of universal safety is made.

### DIFF
--- a/IntelRaptorLakeOcMailbox.p
+++ b/IntelRaptorLakeOcMailbox.p
@@ -1,0 +1,262 @@
+//  IntelRaptorLakeOcMailbox - A narrow PawnIO module for MSR 0x150 OC-mailbox
+//  voltage offsets on the tested Raptor Lake-S desktop model family.
+//
+//  Copyright (C) 2026  RaptorLakeVF contributors
+//
+//  This library is free software; you can redistribute it and/or
+//  modify it under the terms of the GNU Lesser General Public
+//  License as published by the Free Software Foundation; either
+//  version 2.1 of the License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+//  Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public
+//  License along with this library; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+//  SPDX-License-Identifier: LGPL-2.1-or-later
+//
+//  STATUS: SOURCE ONLY. This module is NOT built, NOT signed, NOT loaded, and is
+//  NOT loadable in this repository. It is prepared for review and (later) for
+//  submission to the PawnIO maintainers. A locally built unsigned artifact is
+//  REVIEW ONLY and must never be placed where the application could discover it.
+//  The REAL build toolchain is the pinned PawnIO.Modules compiler (pawncc
+//  4.1.7152 from the `_pawn/` RPMs); the C file in this directory is an
+//  AUXILIARY syntax mirror only (gcc stub) and is NOT the module.
+//
+//  PRIVILEGED SURFACE (the only operations this module performs):
+//    ioctl_query_ia_core_offset()      -> legacy plane-offset QUERY, domain 0
+//    ioctl_query_ring_offset()         -> legacy plane-offset QUERY, domain 2
+//    ioctl_set_ia_core_neg_offset()    -> legacy plane-offset SET, domain 0
+//    ioctl_set_ring_neg_offset()       -> legacy plane-offset SET, domain 2
+//
+//  HARD INVARIANTS:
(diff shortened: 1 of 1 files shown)


namazso:
> An application that only needs the OC mailbox should not hold a module capable of other privileged writes. 

Why? Whoever has rights to load this can just load IntelMSR the same. If you intend to expose the full functionality of a module to unprivileged users, that's a you problem, do the filtering in whatever component does that.

PR #106, opened 2026-08-28
Add LpcECPMC module for the 0x68/0x6C EC PMC interface

lhzlhz419:
## Summary

Add an `LpcECPMC` module providing byte access to an EC PMC/OEM host
interface through:

- `0x68`: data port
- `0x6C`: status/command port

The API mirrors `LpcACPIEC`, while access is strictly restricted to these
two ports.

## Background

This interface layout is present in Lenovo and various other embedded
controller implementations. It is not exclusive to a single manufacturer,
although the commands accepted through the interface may remain vendor- and
firmware-specific.

The module therefore provides only the transport mechanism and does not
contain device-specific commands or protocol assumptions.

## Safety

- Reads and writes are restricted to ports `0x68` and `0x6C`.
- Write values outside the byte range `0x00–0xFF` are rejected with
  `STATUS_INVALID_PARAMETER`.
- Other ports are rejected with `STATUS_ACCESS_DENIED`.
- The IOCTL documentation instructs callers to acquire the
  `\BaseNamedObjects\Access_EC` mutant before accessing the interface.

## References

- ITE embedded-controller documentation describes additional Power Management
  Channels with command/status and data registers.
- Public EC implementations use `0x68` as the OEM data port and `0x6C` as the
  OEM status/command port.

## Testing

- Verified against the existing `LpcACPIEC` IOCTL structure.
- Passed `git diff --check`.

### DIFF
--- a/LpcECPMC.p
+++ b/LpcECPMC.p
@@ -1,0 +1,72 @@
+//  PawnIO Modules - Modules for various hardware to be used with PawnIO.
+//  Copyright (C) 2026  lhzlhz419
+//
+//  This library is free software; you can redistribute it and/or
+//  modify it under the terms of the GNU Lesser General Public
+//  License as published by the Free Software Foundation; either
+//  version 2.1 of the License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+//  Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public
+//  License along with this library; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+//  SPDX-License-Identifier: LGPL-2.1-or-later
+
+#include <pawnio.inc>
+
+// This EC PMC/OEM host-interface layout is present in various embedded
+// controller implementations, including Lenovo systems. Commands sent through
+// the interface remain vendor- and firmware-specific.
+
+is_port_allowed(port) {
+    return port == 0x68 || port == 0x6C;
+}
+
+/// Read byte from the EC PMC/OEM host interface.
+///
+/// @param in [0] = Port
+/// @param in_size Must be 1
+/// @param out [0] = Value read
+/// @param out_size Must be 1
+/// @return An NTSTATUS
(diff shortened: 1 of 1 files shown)


namazso:
I changed this to draft, please test it on the unrestricted driver (you'll have to use 2.1.0 or your own builds because 2.2.0 accidentally shipped with signature verification enabled on the unrestricted driver as well) before marking it as ready.

lhzlhz419:
> I changed this to draft, please test it on the unrestricted driver (you'll have to use 2.1.0 or your own builds because 2.2.0 accidentally shipped with signature verification enabled on the unrestricted driver as well) before marking it as ready.

Tested successfully on the following system:

- Lenovo 21R0
- BIOS R2CN57WW
- Windows build 26200.8894 (25H2)
- x64

Validation performed:

- Compiled `LpcECPMC.p` successfully with Pawn compiler 4.1.7152 using the repository CI flags.
- Built PawnIO 2.1.0 from the official source with `PAWNIO_UNRESTRICTED=ON` for x64.
- Loaded and unloaded the compiled module successfully.
- Confirmed that a read from disallowed port `0x67` is rejected with `0x80070005`.
- Confirmed that writing value `0x100` is rejected with `0x80070057`.

The allowed read and write paths were tested with a known read-only EC transaction. Each RPM byte was obtained using the following sequence:

1. Read the status register from port `0x6C` and wait for IBF to clear.
2. Write command `0x7E` to port `0x6C`.
3. Write the EC address high and low bytes to data port `0x68`, waiting for IBF to clear after each byte.
4. Poll port `0x6C` until OBF is set.
5. Read the returned byte from data port `0x68`.

This sequence was repeated for four EC offsets. The returned low/high byte pairs were combined into the two fan RPM values. Therefore the RPM samples exercise all four operations exposed by the module:

- read from `0x6C`
- write to `0x6C`
- write to `0x68`
- read from `0x68`

Three consecutive samples were returned successfully:

- FAN1 2513 RPM (`D1 09`), FAN2 2301 RPM (`FD 08`)
- FAN1 2507 RPM (`CB 09`), FAN2 2303 RPM (`FF 08`)
- FAN1 2510 RPM (`CE 09`), FAN2 2321 RPM (`11 09`)

For example, the first FAN1 value was reconstructed as `0x09D1 = 2513 RPM`.

The RPM offsets and command semantics belong to the tested firmware and are not implemented or assumed by `LpcECPMC`; they were used only to verify the module's restricted byte I/O transport.

No fan-control or full-speed command was issued during the test.

The resulting AMX SHA-256 is:

`473F90DBA6502271AF445E57652F46DCDAABC97EBC66E77738038FCD65C08644`

Issue #72, opened 2026-07-04
Module fails to load with error 80070057, even with unmodified Echo.p


ap1kazakov-byte:
# PawnIO module fails to load with `80070057` (E_INVALIDARG) — even official unmodified `Echo.p`

## Environment

- **Laptop:** Honor MagicBook X16 (2026), motherboard identified as `BRG-X-PCB`
- **CPU:** Intel Core Ultra 5 125H (x64)
- **OS:** Windows 10/11, build `10.0.26200.0` (reported by `[System.Environment]::OSVersion.Version`)
- **Note:** This is a modified/"debloated" third-party Windows build (not a stock Microsoft ISO). I flag this transparently as a possible contributing factor, though core OS components (services, drivers, PowerShell) otherwise function normally.
- **PawnIO version:** 2.2.0.0, installed in **unrestricted** edition
- **PawnIOUtil.exe / PawnIOLib.dll:** same 2.2.0.0 package, from `PawnIO_setup.exe -install -silent -unrestricted`
- **Pawn compiler:** tested with both `4.1.7487` (from compuphase.com self-extracting installer) and `4.1.7152` (matching the exact version pinned in `PawnIO.Modules`' own CI, `_pawn/pawn-4.1.7152-1.el9.x86_64.rpm`)
- **Admin rights:** confirmed (`WindowsPrincipal.IsInRole(Administrator) == True`)
- **Test signing:** enabled and confirmed active (`bcdedit /enum` shows `testsigning: Yes`)

## Goal

Writing a custom PawnIO module to read/write a proprietary EC "mailbox" protocol on this laptop (physical MMIO region at `0xFE0B0400`, reverse-engineered from the machine's DSDT) for a Fan Control plugin. Before debugging my own module, I isolated the problem down to the **official, completely unmodified `Echo.p`** sample from the `PawnIO.Modules` repository — which also fails, confirming the issue is environmental, not in my code.

## Steps taken

1. Installed PawnIO in **unrestricted** mode:
   ```
   PawnIO_setup.exe -install -silent -unrestricted
   ```
   Confirmed via `Get-Service PawnIO` → `Running`.

2. Confirmed driver file signature (informational — I understand this doesn't distinguish official vs. unrestricted builds, both are signed the same way):
   ```
   Get-AuthenticodeSignature ...\PawnIO.sys
   → Status: Valid
   → Subject: E=admin@namazso.eu, CN=namazso.eu, O=namazso, L=Debrecen, C=HU
[...]

PR #107, opened 2026-08-28
Add PawnIO backend for legacy system speaker (aka PC speaker) access

GeniusPilot2016:
Add the PcSpeaker.p PawnIO module for NeoBleeper to access the legacy system speaker (aka PC Speaker) hardware.

The module provides the hardware-level functionality required for generating beeps, checking whether system speaker hardware is present, and handling related system speaker responsibilities.

Replace the previous InpOutx64-based implementation due to compatibility concerns following the Windows 11 August 2026 update and the possibility of the InpOutx64 driver being blocked by the September 2026 update and later Windows releases.

This moves NeoBleeper's system speaker access to PawnIO and avoids depending on the InpOutx64 driver for direct hardware I/O.

This module compiles successfully but has not yet been tested on physical hardware. Testing is currently blocked by PawnIO Setup 2.2.0 issue #72.The implementation requires hardware validation before merge.

### DIFF
--- a/PcSpeaker.p
+++ b/PcSpeaker.p
@@ -1,0 +1,343 @@
+//  PawnIO Modules - Modules for various hardware to be used with PawnIO.
+//  Copyright (C) 2026  M-Control Contributors
+//
+//  This library is free software; you can redistribute it and/or
+//  modify it under the terms of the GNU Lesser General Public
+//  License as published by the Free Software Foundation; either
+//  version 2.1 of the License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+//  Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public
+//  License along with this library; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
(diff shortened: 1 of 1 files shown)


namazso:
Untested, slop, and I don't want to hijack the timer hardware, it's a resource owned by Windows. The correct way to beep is `HalMakeBeep`, none of this extra stuff.

Plus someone already made and signed BEEPx / BEEPx64, you can just use it as-is: https://dose.0wnz.at/waldbauer.com/

GeniusPilot2016:
> Untested, slop, and I don't want to hijack the timer hardware, it's a resource owned by Windows. The correct way to beep is `HalMakeBeep`, none of this extra stuff.
> 
> Plus someone already made and signed BEEPx / BEEPx64, you can just use it as-is: https://dose.0wnz.at/waldbauer.com/

I was used that driver (latest version) and that driver has signature issues that its signature can't be verified in Windows 10 21H2 and above and HalMakeBeep is not user mode method that's a ring0 method that can't be used in applications without driver.

namazso:
> HalMakeBeep is not user mode method that's a ring0 method that can't be used in applications without driver.

Which is exactly what `get_proc_address` and `invoke` is for.

GeniusPilot2016:
> > HalMakeBeep is not user mode method that's a ring0 method that can't be used in applications without driver.
> 
> Which is exactly what `get_proc_address` and `invoke` is for.

Hmmmm... I remembered now, I used the BeepXP64's both direct and HalMakeBeep-based version, but produced no sound in HalMakeBeep version and direct version one produced and the BeepXP64 has driver signature issue in both of them.

GeniusPilot2016:
Now, I tested in older version of PawnIO and it works without any issues.

PR #108, opened 2026-08-30
Add PawnIO module for legacy PC speaker access

GeniusPilot2016:
Add PcSpeaker.p to provide NeoBleeper with direct access to the legacy PC speaker on modern Windows systems.

The module programs the legacy speaker interface to:

- Generate tones
- Change tone frequency
- Start and stop continuous tones
- Check whether the PC-speaker-related hardware interface is present and responsive

The hardware check covers PIT channel 2 and System Control Port B. It does not claim to detect whether a physical speaker cone is connected or whether its output is audible.

This implementation replaces NeoBleeper's previous InpOutx64-based backend because of current and upcoming Windows compatibility issues. Microsoft has confirmed deployment of a block that prevents the inpoutx64 driver from loading on affected Windows 11 devices. Microsoft also states that this block will be included in the September 2026 Windows security updates and subsequent releases.

Source:
https://learn.microsoft.com/en-us/windows/release-health/resolved-issues-windows-11-25h2#3731msgdesc

Existing alternatives were unsuitable for this use case. BEEPx64 has driver-signing compatibility problems on current Windows versions, and HalMakeBeep did not reliably produce physical PC-speaker output during prior testing. PawnIO provides a maintained mechanism for performing the required hardware I/O without depending on InpOutx64.

The module compiled successfully and was tested with PawnIO 2.0.1 on:

- A desktop with a Gigabyte B450M H motherboard and a physical system speaker
- A Huawei MateBook D16 laptop

Tone generation, frequency changes, and start/stop operations worked without observable errors or latency on both systems.

PawnIO 2.0.1 was used because the unrestricted PawnIO 2.2.0 build rejects unofficial modules with STATUS_INVALID_PARAMETER. The unrestricted build appears to behave like the standard build, which accepts only officially signed modules.

### DIFF
--- a/PcSpeaker.p
+++ b/PcSpeaker.p
@@ -1,0 +1,368 @@
+//  PawnIO Modules - Modules for various hardware to be used with PawnIO.
+//  Copyright (C) 2026 GeniusPilot2016
+//
+//  This library is free software; you can redistribute it and/or
+//  modify it under the terms of the GNU Lesser General Public
+//  License as published by the Free Software Foundation; either
+//  version 2.1 of the License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+//  Lesser General Public License for more details.
+//
+//  You should have received a copy of the GNU Lesser General Public
+//  License along with this library; if not, write to the Free Software
+//  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+//  SPDX-License-Identifier: LGPL-2.1-or-later
+
+#include <pawnio.inc>
+
+// Legacy system speaker module.
+//
+// This module is intended only to generate tones through the legacy system
+// speaker and to check whether the system-speaker-related hardware interface is
+// present and responsive. It uses system timer channel 2, which is the channel
+// associated with the system speaker; it does not access the timer channels
+// used for Windows timekeeping.
+//
+// The hardware check covers system control port B and the system timer channel
+// 2 feedback path. It cannot determine whether a physical speaker cone is
+// connected or whether a generated tone is audible.
+//
+// System timer channel 2 generates the square wave. Bits 0 and 1 of system control
+// port B connect that signal to the speaker. Other bits in port 0x61 are
+// preserved because firmware and other legacy devices may use them.
+//
+// The reason behind creating this module is HalMakeBeep is a ring0 function and cannot be called from user mode and if called from kernel mode, no sound is produced in modern systems.
+//
+// Also, there's a driver called Beep.sys that can be used to make a beep, but it uses ordinary speaker instead of the system speaker
+// and there was a replacement called BeepXP64 but it has driver signature issues (the signature that can't be verified) and its download link is dead that only "404 - page not found" is shown when you click on the original download link. 
+// So, this module is created to make a beep sound using the system speaker.
+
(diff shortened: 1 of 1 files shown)


namazso:
Please test if HalMakeBeep works instead, it’d be much preferable as it correctly locks the mutex for the device

GeniusPilot2016:
> Please test if HalMakeBeep works instead, it’d be much preferable as it correctly locks the mutex for the device

I already tested HalMakeBeep first (via BeepXP64’s HalMakeBeep path).
On my systems it produced no audible output (including with driver signature enforcement disabled), while direct PC-speaker I/O worked reliably.

Since PawnIO.Modules is for direct hardware I/O modules, this PR intentionally uses direct access.

Please confirm which direction you want:

keep direct access (works on my tested systems), or
require HalMakeBeep/mutex-correct path even if it yields no sound on these systems.

GeniusPilot2016:
> Please test if HalMakeBeep works instead, it’d be much preferable as it correctly locks the mutex for the device

image


I built a side-by-side test app to compare both methods on the same system.
Result: Direct access works, while HalMakeBeep failed with: “Bir sisteme bağlı bir aygıt çalışmıyor. (0x8007001F)” — English: “A device attached to the system is not functioning. (0x8007001F)”.

So on this machine, HalMakeBeep is non-functional for audible PC-speaker output, while direct hardware access is functional.
Given PawnIO’s purpose (direct hardware I/O), this PR uses the only working path in my tests.

namazso:
I see, interesting. I checked the code and I see you have quite some processing and gating going on in the module. Considering your goal of playing back actual audio-ish stuff on it, I question your approach. Have you tried any higher quality methods?

What I have in mind is that instead of using the square wave output at a certain frequency, you could instead just pick a (high) carrier frequency, and use port 0x61 bit 1 with a pulse density modulator. I imagine the best pipeline would be like this:

1. resample audio to carrier frequency
2. high-pass since we don't want DC-ish components
3. equalize (optional, a speaker could be calibrated with frequency sweep)
4. low-pass
5. dither
6. sigma-delta modulator into 0x61 bit 1

Is there anything I don't see here? If this is possible we should almost certainly just expose all the toys instead of just a frequency setting interface.

GeniusPilot2016:
> I see, interesting. I checked the code and I see you have quite some processing and gating going on in the module. Considering your goal of playing back actual audio-ish stuff on it, I question your approach. Have you tried any higher quality methods?
> 
> What I have in mind is that instead of using the square wave output at a certain frequency, you could instead just pick a (high) carrier frequency, and use port 0x61 bit 1 with a pulse density modulator. I imagine the best pipeline would be like this:
> 
> 1. resample audio to carrier frequency
> 2. high-pass since we don't want DC-ish components
> 3. equalize (optional, a speaker could be calibrated with frequency sweep)
> 4. low-pass
> 5. dither
> 6. sigma-delta modulator into 0x61 bit 1
> 
> Is there anything I don't see here? If this is possible we should almost certainly just expose all the toys instead of just a frequency setting interface.

Thanks, this is a very good suggestion.

The processing/gating in this module is intentionally not aimed at high-quality audio synthesis. In this PR, those paths are mainly for:

1.  Functionality testing / hardware validation (confirming PIT ch2 + port 0x61 path is responsive), and
a system-speaker wake-up workaround needed on some machines before normal tone output behaves reliably.
2.  The actual “audio-ish” rendering strategy (including modulator choices) is handled at a higher level in NeoBleeper, not inside this low-level module and only have beeping method, start sound and stop sound methods as core methods to produce sound.

So the module API currently focuses on deterministic low-level control (set/start/stop/check), while the more advanced DSP/modulation pipeline is expected to live in NeoBleeper where it can evolve independently.

That said, I agree your carrier + PDM/sigma-delta direction is valid and interesting for better fidelity. We can expose additional low-level primitives later if we decide NeoBleeper needs tighter real-time control over that path.

PR #110, opened 2026-08-30
Add RadeonSMU module

miklebel:
## RadeonSMU: SMU mailbox access for AMD Radeon (RDNA) GPUs

Replaces WinRing0/InpOut32-style raw physical MMIO in AMD GPU tuning
software. Rather than exposing a physical read/write primitive, the module
owns the SMN index/data sequence itself and allowlists the SMN addresses a
caller may reach, so the exposed surface is the MP1 message mailbox and
nothing else.

### Why this rather than a first-party driver

There is no first-party interface for this. AMD's Windows display driver
exposes Overdrive through D3DKMT escapes, but not the SMU message mailbox,
and tools that need the mailbox today ship WinRing0 or InpOut32 — i.e. an
unrestricted physical read/write primitive available to any process that can
open the device. This module is a strictly narrower replacement for that
specific pattern, which the contribution guidelines list as preferred.

### Surface

| ioctl | in → out | scope |
|---|---|---|
| `ioctl_read_smn` | `[smn]` → `[value]` | MP1 C2PMSG only |
| `ioctl_write_smn` | `[smn, value]` → `[status]` | MP1 C2PMSG only |
| `ioctl_read_metrics` | `[pa]` → 65 dwords | VRAM aperture only, read-only |
| `ioctl_get_bounds` | `[_]` → `[ready, reg_bar, reg_size, vram_bar, vram_size]` | — |

All four are `DEFINE_IOCTL_SIZED` — no caller-sized arrays.

**SMN allowlist.** `0x03B10900..0x03B10AFF` — the 128-register MP1 C2PMSG
file, dword-aligned. That window is the SMU message mailbox: MSG
(C2PMSG_66), PARAM (C2PMSG_82), RESP (C2PMSG_90), and the addr_hi/lo pair
(C2PMSG_80/81) that PMFW populates for `TransferTableSmu2DramWithAddr`.
Every other SMN address is refused.

**No message-id gating.** Every PPSMC message is a write of an id to the same
MSG register, so allowlisting the registers rather than the messages keeps
present and future firmware messages working without widening the surface.
This mirrors `RyzenSMU`, which likewise bounds SMN ranges and accepts any
message id.

**VRAM is read-only, and only for one buffer.** `ioctl_read_metrics` reads
the `SmuMetrics_t` DMA buffer that the AMD display driver itself allocated
and programmed; the module neither allocates it nor writes to it. The caller
recovers its address from C2PMSG_80/81, and the module still bounds-checks
that address against the probed BAR0 aperture rather than trusting it. There
is no write path to the framebuffer at all.

### Device discovery

The module finds the AMD display device itself by PCI scan; the caller never
supplies a bus/device or a physical address for register access. Selection is
by **largest VRAM aperture**, not enumeration order, so an AMD iGPU that
enumerates first (e.g. Raphael, 256 MB carved from system RAM) cannot shadow
a discrete card. `ioctl_get_bounds` lets a caller confirm which device was
chosen and observe the aperture across ReBAR states.

Compatibility checks before anything is trusted: vendor `0x1002`, PCI class
`03:00` (display/VGA), BAR0 must be a 64-bit memory BAR before `0x14` is
treated as its high half, BAR5 must be a memory BAR, and the register
aperture must be large enough to contain PCIE_INDEX2/DATA2. On any failure —
non-x64, no AMD VGA device, an unreadable or implausible BAR — `main()`
returns `STATUS_NOT_SUPPORTED` and the module does not load at all, so a
caller finds out at load time rather than through a run of
`STATUS_ACCESS_DENIED`.

### Bounds-check notes

- **No overflow.** Cells are signed 64-bit under `-C64`, so the naive
  `pa + len <= base + size` can wrap for a `pa` near 2^63 and wrongly return
  true. `in_window()` never adds to `pa`: it checks `pa >= base` (which also
  rejects a negative `pa`), derives the offset by subtraction, and compares
  `len` against the remaining window, which cannot overflow.
- **Both probed sizes are sanity-capped** — `REG_SIZE_MAX` (16 MB) for BAR5
  with a fallback, `VRAM_SIZE_MAX` (64 GB) for BAR0. An implausible read-back
  skips the device rather than widening a window.
- **Alignment is enforced** on both the SMN address and the metrics address.
- The BAR size probe uses the standard write-all-ones / read-mask / restore
  sequence. Memory decode is **not** disabled during it — PCI config access
  is PASSIVE_LEVEL and cannot be IRQL-protected — and each BAR is restored
  immediately. Happy to change this if you would rather it were handled
  differently.

### Scope, stated plainly

The register BAR contains the SMN index/data pair, which is an aperture onto
the GPU's internal SMN bus. Bounding SMN addresses to the C2PMSG file is what
confines this module to the mailbox; without that allowlist, BAR5 access
would reach the whole SMN space. That is why the allowlist, not the BAR
bound, is the real boundary here.

### Reference

- Linux `amdgpu`: `drivers/gpu/drm/amd/pm/swsmu/smu_cmn.c` — mailbox protocol
  (poll RESP, clear RESP, write PARAM, write MSG, poll RESP).
- `amdgpu_device.c` — PCIE_INDEX2/PCIE_DATA2 indirect SMN access.
- `smu_v14_0.c` / `smu_v13_0.c` — `mmMP1_SMN_C2PMSG_66/82/90` numbering.
- `smu14_driver_if_v14_0.h` — Navi 4x `SmuMetrics_t` (260 bytes = 65 dwords).

### Testing

Tested on an **AMD RX 9060 XT** (Navi 44, PCI `1002:7590`, PMFW 0.102.70.0),
with an NVIDIA card also present in the machine, under a locally built
`PAWNIO_UNRESTRICTED` driver (2.2.0 source) so the unsigned module could be
loaded. Compiles with **no warnings** under the exact CI invocation
(`pawncc RadeonSMU.p -iinclude -C64 -;+ -(+ -p`), against this repository's
own `include/`.

Full run against the submitted source (`0x80070005` = `ERROR_ACCESS_DENIED`):

```
bounds     ready=1  reg_bar=0xDF000000 size 0x80000
                    vram_bar=0xF000000000 size 0x400000000 (16 GiB, ReBAR on)

allowlist  read C2PMSG_66 (0x3B10A08)  hr=0x00000000 value=0x00000052  allowed
           read C2PMSG_90 (0x3B10A68)  hr=0x00000000 value=0x00000001  allowed
           read SMN 0x0                hr=0x80070005  DENIED (outside mailbox)
           read one dword BELOW base   hr=0x80070005  DENIED
           read one dword ABOVE end    hr=0x80070005  DENIED
           read unaligned              hr=0x80070005  DENIED (alignment)
           write SMN 0x21000           hr=0x80070005  DENIED (outside mailbox)

mailbox    GetSmuVersion (msg 0x02) via ioctl_write_smn / ioctl_read_smn
           RESP        = 0x00000001  (OK)
           SMU version = 0x00664600  (0.102.70.0)

metrics    C2PMSG_80/81 = 0x00000083 / 0xF6DC6000 -> pmfw 0x83F6DC6000
           translated   = 0xF3F6DC6000  (vram_bar + 0x3F6DC6000)
           ioctl_read_metrics          hr=0x00000000  allowed, 40/65 non-zero
           read metrics past aperture  hr=0x80070005  DENIED
           read metrics at RAM 0x1000  hr=0x80070005  DENIED
```

The SMU version returned through the allowlisted ioctls matches what the AMD
driver reports, so the mailbox protocol works end to end inside the
allowlist. The `0x7FFFFFFFFFFFFFF0` overflow probe is likewise refused; with
a naive `pa + len <= base + size` check it wrapped and was *allowed*.

Aperture tracking was exercised in both ReBAR states. With ReBAR **off** the
same code re-probes `vram_bar=0xF820000000 / 0x10000000` (256 MB) and the
metrics buffer moves with it to `0x800EFC6000 -> 0xF82EFC6000`, ~16 MB below
the top of the smaller window; with ReBAR **on**, the values above. Nothing
is cached across loads — the PCI scan and both BAR size probes re-run every
time. (The ReBAR-off pass was run against the immediately preceding
revision, which differs only in `main()` returning `STATUS_NOT_SUPPORTED`
when no GPU is found; the discovery and bounds code is identical.)

### Consumer

[Adrenalift](https://github.com/miklebel/adrenalift) — a closed-source AMD
RDNA tuning utility. Its current public release (0.8.0) uses **InpOut32**,
i.e. exactly the unrestricted physical read/write primitive this module is
meant to replace. An internal build has already been ported to the four
ioctls above and has no other hardware path: the physical-MMIO surface was
removed from the host rather than left as a fallback.

That build cannot be released until the module loads under the stock signed
driver, so this is not a speculative submission — it is the blocker on
dropping InpOut32 from a shipping product.

### DIFF
--- a/RadeonSMU.p
+++ b/RadeonSMU.p
@@ -18,7 +18,7 @@
 //  SPDX-License-Identifier: LGPL-2.1-or-later
 
 /*
- * RadeonSMU - SMU mailbox access for AMD Radeon (RDNA) GPUs.
+ * RadeonSMU - SMU mailbox access for AMD Radeon RDNA4 (Navi 4x) GPUs.
  *
  * Replaces WinRing0/InpOut32-style raw physical MMIO for AMD GPU tuning
  * software. Rather than handing out a physical read/write primitive, this
@@ -36,6 +36,10 @@
  * discrete card. `ioctl_get_bounds` lets the caller confirm which device
  * was chosen.
  *
+ * Discovery is read-only: the VRAM aperture size comes from the Resizable
+ * BAR capability (`rebar_current_size`), not a write-all-ones probe. Only
+ * Navi 4x is bound (`is_supported_gpu`), the one family this has run on.
+ *
  * Exposed surface
  * ---------------
  *   - SMN reads/writes restricted to the MP1 C2PMSG register file
@@ -45,8 +49,17 @@
  *     populates for TransferTableSmu2DramWithAddr.
  *   - One fixed-size read of the SmuMetrics_t DMA buffer, bounded to the
  *     GPU's own VRAM aperture.
- *   - Nothing else. No system RAM, no other devices, no writes to the
- *     framebuffer, and no SMN address outside the mailbox window.
+ *   - Nothing else. The module itself touches no system RAM, no other
+ *     device, and no SMN address outside the mailbox window, and it never
+ *     writes to the framebuffer.
+ *
+ * Two caveats, both shared with RyzenSMU. C2PMSG_80/81 are writable, so a
+ * caller can aim the firmware's own DMA anywhere via
+ * TransferTableSmu2DramWithAddr; that transfer is the SMU's, not ours, and
+ * no bound here applies to it. And PCIE_INDEX2/DATA2 is the same pair the
+ * AMD display driver uses (Linux: `adev->pcie_idx_lock`); we cannot take
+ * that lock, so callers should serialise via the
+ * "\BaseNamedObjects\Access_PCI" mutant.
  *
  * The mailbox window is deliberately not gated on message id: every
  * PPSMC message is a write of an id to the same MSG register, so
@@ -63,6 +76,9 @@
  *   - MP1 C2PMSG register numbering: `smu_v14_0.c` / `smu_v13_0.c`,
  *     `mmMP1_SMN_C2PMSG_66/82/90`.
  *   - Navi 4x SmuMetrics_t layout: `smu14_driver_if_v14_0.h`.
+ *   - Resizable BAR capability: PCI Express Base Specification, "Resizable
+ *     BAR Extended Capability"; Linux `drivers/pci/pci.c`
+ *     `pci_rebar_find_pos` / `pci_rebar_get_current_size`.
  */
 
 #include <pawnio.inc>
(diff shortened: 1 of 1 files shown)


DevTechProfile:
I will add RDNA2 and 3 support based on the Linux implementation and then test it on 

1. RX 6800 XT
2. RX 7900 XTX 
3. RX 9070 XT

against HWiNFO as reference.

DevTechProfile:
RDNA3 already working...

image

DevTechProfile:
RDNA3 support is now validated. I ported the ABI 2 implementation, including fixed-size SMU13.0.0/13.0.10 and SMU13.0.7 metrics reads, in [4975b24](https://github.com/DevTechProfile/CX.PawnIO.Modules/commit/4975b2438c42e25a46afedea79cba71b24e6aafd). I cannot push to this PR's head fork, so please cherry-pick that commit.

miklebel:
@DevTechProfile Thank you for digging into this! I don't have anything but navi 44 so RDNA2/3 is not something I can try alone.

Rather than cherry picking could you open a PR against that branch directly? https://github.com/miklebel/PawnIO.Modules/compare/radeonsmu...DevTechProfile:CX.PawnIO.Modules:radeonsmu-rdna3?expand=1

You kept `in_window()` in the overflow-safe form rather than simplifying it, and `resolve_metrics_buffer` validating the firmware address is better than what I had. 
Happy to take that.

A few things:
- The header says "a caller cannot choose a physical address". `C2PMSG_80/81` sit inside the range `ioctl_write_smn` allows, so anyone can write them and steer the derived address, it's still bounded to the aperture so nothing escapes and honestly the legacy `ioctl_read_metrics` already takes an address directly, it grants nothing new.
- The docs got stripped out. Could you restore those?
- RDNA2/RDNA3  genuine question, since you have the cards and I don't. Your own header says Navi 31 was verified to leave `C2PMSG_80/81` at zero. What did it look like on your side? did you get a real table back, or was it validated a different way?
- could you confirm the table sizes against the Linux headers?
- And 13.0.10 in a couple of doc comments should be 13.0.0. `ioctl_read_metrics_rdna3` and `_rdna3_0` are identical, worth keeping just one unless you want the alias.

DevTechProfile:
Once the RDNA2 support is validated, I will open a PR and look into the points you have mentioned. Just to be safe, I will also test the module with my RX 9070 XT. We also need a direct comparison against HWiNFO!

DevTechProfile:
RDNA2 support working
image

DevTechProfile:
Forget it, that was the ADL fallback. Now we are talking...
image

DevTechProfile:
For RDNA3 there so many sensors, you can't display all of them without scrolling. 
image

image

DevTechProfile:
PawnIO module vs HWiNFO (RDNA3)

image

DevTechProfile:
PawnIO module vs HWiNFO (RDNA4)

image

DevTechProfile:
The requested follow-up PR is now available at https://github.com/miklebel/PawnIO.Modules/pull/1. It contains only `RadeonSMU.p` and incorporates the review feedback, including bounded RDNA2, RDNA3, and RDNA4 telemetry paths.

DevTechProfile:
Module consumer code: https://github.com/CXWorld/CapFrameX/commit/cbdb98b13522fcad0954898e4c5fc35697215813

miklebel:
Pushed a revision on top of the original submission. Summary of what changed:
- Discovery is now read-only. The BAR0 size comes from the Resizable BAR capability's control register instead of the write-all-ones probe, and the BAR5 probe is gone (its result was never used). Nothing in the module writes to config space any more.
- Unassigned BARs are rejected. A BAR with no resources still reads its type bits, so the old code accepted a zero base and bounded every VRAM read against low system RAM. Both apertures must now be strictly positive, which is also what keeps in_window()'s overflow argument sound.
- BAR5 must be a 32-bit memory BAR. The old 64-bit branch read 0x28, which is the Cardbus CIS Pointer.
- Device gate on Navi 4x. On an APU the C2PMSG window this module allowlists is the CPU's SMU mailbox (the addresses RyzenSMU uses on Renoir through StrixHalo all fall inside it), so it now refuses to bind anything but the family it has been tested on. Ranges for RDNA2/3 are cited in a comment for whoever widens it with test evidence.
- Read statuses checked throughout; the header now states plainly what the mailbox can still be asked to do via C2PMSG_80/81 and that the index/data pair is shared with the display driver. 

ioctl ABI is unchanged. Tested on an RX 9060 XT: the ReBAR read returns the same 16 GB aperture the old probe measured, the mailbox is live, and out-of-window SMN addresses are denied.

miklebel:
@DevTechProfile - credit where it's due: the revision I just pushed exists because of your PR. Reviewing your discovery rewrite is what made me notice the BAR probe was writing to a live GPU and that an unassigned BAR sailed through with a zero base. I'd have shipped both. Your RDNA2/3 validation and the HWiNFO comparisons are also more coverage than this module has ever had, and CapFrameX using it is exactly the kind of consumer it was written for.

I've asked you to take the RDNA2/3 telemetry work to upstream directly once this lands rather than through my branch, and I want to be clear that's about review order, not about the work. I don't have the hardware to stand behind it, and you shouldn't need me in the middle. I'll review the rebase before you open it if that helps, and I'm happy to be pinged on it here.

DevTechProfile:
I think RDNA2 and RDNA3 support should be included before #110 is merged. The implementations have already been
validated on an RX 6800 XT and RX 7900 XTX and compared against HWiNFO, so leaving the initial module Navi 4x-only
would unnecessarily ship a much narrower interface.

I agree that the branch should be rebased onto your revised discovery code rather than merged as-is. I can also
address the polling, address-translation, and citation issues you identified and split the generation support into
focused commits to keep the review manageable.

My preference would therefore be to keep #110 open until the tested RDNA2/RDNA3 paths are integrated, instead of
merging a Navi 4x-only ABI and extending it immediately afterward.

